PR #4, opened 2019-12-05
chore(build): set up github actions

lburgazzoli:


orpiske:
LGTM. The tests take about 10 minutes to run per iteration, so if we don't have any timeout on the actions, this should be fine.

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,37 @@
...
+      matrix:
+        java: [ '1.8', '11' ]
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK ${{ matrix.java }}
+      uses: actions/setup-java@v1
+      with:
+        java-version: ${{ matrix.java }}
+    - name: mvn verify
+      run: ./mvnw -V -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn clean install
```
zregvart: [With Maven 3.6.1 onward](https://maven.apache.org/docs/3.6.1/release-notes.html#User_visible_Changes) there is a  `--no-transfer-progress` option

lburgazzoli: Yep, but the project seems to be 3.6.0 now, right ?

oscerd:
with this configuration do we have docker at least at version 17.0? This is needed for integration test

lburgazzoli:
it should, lets try

PR #6, opened 2019-12-05
Fix Sink File test

orpiske:
Prevent the file test from failing when the test file has been created but no content has been written


oscerd:
Thanks

PR #27, opened 2019-12-09
Fix camel file test

orpiske:
Prevent the file test from failing when the test file has been created but no content has been written


orpiske:
Please ignore.

Issue #13, opened 2019-12-08
Unclean shutdown causing lots of warning messages in the logs


orpiske:
When the test is shutting down, there are a lot of messages like this in the logs (~9k messages like this for a test run). 

`2019-12-08 20:24:49,992 [7c-6916da6c7843] WARN org.apache.kafka.clients.NetworkClient - [Producer clientId=6a23d856-c3d2-40b6-be7c-6916da6c7843] Connection to node 1 (localhost/127.0.0.1:34649) could not be established. Broker may not be available.`

This doesn't really affect anything, so probably low priority, but it seems to be fixed by a separate patch I am working on, so I may refer to this one when I send that one.

orpiske:
Closing this one because it was fixed with the merge on #29.

valdar:
Thanks! If you mention fix #issuenumber in the commit message it should close it automatically, I hope.

PR #29, opened 2019-12-09
Support using an external Kafka for the tests

orpiske:
This decouples the kafka test service from the local container and allows using an
external Kafka instance for testing. It also creates the base work to run the tests
with different different distributions of Kafka (ie.: different Kafka containers, etc)         

As a side effect of this change if also fixes issue #13

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -28,6 +28,12 @@ To run the integration tests it is required to:
 mvn clean verify package
 ----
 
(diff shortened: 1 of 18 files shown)


orpiske:
It's been acting strangely when running via Github actions, so I'll review this.

valdar:
Test passes, in the end, feel free to reopen it if needed.

PR #34, opened 2019-12-10
Support camel type converter

omarsmak:
This shall address #16. It will initialize the default Camel type registry upon starting the connector. To use it, you can add the following to the `.properties` file as an example:
```
transforms=CamelTypeConverterTransformer
transforms.CamelTypeConverterTransformer.type=org.apache.camel.kafkaconnector.transforms.CamelTypeConverterTransform$Value
transforms.CamelTypeConverterTransformer.target.type = java.lang.Float
```
Whereby you can change `org.apache.camel.kafkaconnector.transforms.CamelTypeConverterTransform$Value` to convert the value of a connect record with `org.apache.camel.kafkaconnector.transforms.CamelTypeConverterTransform$Key` to convert the key of a connect record. 
`target.type` is the full qualified java name for the type you want to convert to. 

This transformer will try its best to populate the correct schema for the corresponding type. However, in case the type of the value is unknown, it will by default included it as `BYTES` schema. 

P/S: There are maybe some checkstyle errors as I think the checksyle profile is not included yet?

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -97,6 +97,11 @@
             <artifactId>connect-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-transforms</artifactId>
+            <scope>provided</scope>
+        </dependency>
 
         <!-- Test -->
         <dependency>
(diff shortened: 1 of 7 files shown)


oscerd:
This is really nice. Thanks a lot!

oscerd:
Yes, probably there are checkstyle violations, but we'll fix them once we add the needed bits

omarsmak:
No problems. I will try to improve on it as well :)

PR #48, opened 2019-12-13
Add checkstyle profile 

omarsmak:
Per titile. Basically I took over the buildtool module from camel main project and removed some stuff related to groovy which I think are not important for this project. However, my gut feelings tell me there are more stuff that we can remove which are not important for this project, thus I'd appreciate for feedback on this.


oscerd:
Thank you. Lets have a look at violations now.

omarsmak:
yeah, there are many now :D

PR #53, opened 2019-12-19
Added support for testing the Kafka service using Strimzi

orpiske:
Also updates the default Kafka service to use the containers from the Strimzi project

### DIFF
--- a/tests/src/test/java/org/apache/camel/kafkaconnector/services/kafka/StrimziContainer.java
+++ b/tests/src/test/java/org/apache/camel/kafkaconnector/services/kafka/StrimziContainer.java
@@ -1,0 +1,46 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


orpiske:
The docker-compose file is basically the one from the quarkus quickstarts [1], with an upgraded version of the containers. It is licensed as Apache 2. Do we have to do anything here in terms of attribution?

1. https://github.com/quarkusio/quarkus-quickstarts/blob/master/kafka-streams-quickstart/docker-compose-local.yaml#L16-L30

valdar:
Mmm I'd stay away from compose if possible, isn't there an easy way to run those 2 containers by hand?

orpiske:
> Mmm I'd stay away from compose if possible, isn't there an easy way to run those 2 containers by hand?

I think there is. I actually considered doing this at first, but ended up going for the Docker Compose solution since it would be simpler and require less code. I don't think doing this orchestration by hand would be significantly more complex, tough. Probably just a bit more code. 

Let me get back to the code and see how it would look like.

valdar:
I mean don't get me wrong, if it is an effort just stick with the compose solution that is used in other oss projects as well so...

orpiske:
> I mean don't get me wrong, if it is an effort just stick with the compose solution that is used in other oss projects as well so...

That's OK. I think it's a valid point and, since the effort wouldn't be too much, it was worth taking a look and see the result. 

I implemented a simple version of this using only containers: https://github.com/orpiske/camel-kafka-connector/compare/master...orpiske:replace-kafka-2?expand=1 

I still prefer the docker compose approach because it seems cleaner to me, but I don't have a strong preference. If you think that is a better approach, I can definitely clean it up and replace the commits here with those.

valdar:
@oscerd what is your opinion on this one?

oscerd:
I don't like docker compose at all.. but I understand it could be useful in
this case

Il mer 8 gen 2020, 22:48 Andrea Tarocchi <notifications@github.com> ha
scritto:

valdar:
Let's start with the version without docker compose we have here: https://github.com/apache/camel-kafka-connector/pull/66 we can always revert to this approach if needed.

PR #62, opened 2020-01-06
Use artifactId instead of project.name to build colon-less distro paths

ppalaga:
project.name contains colons and thus 

```
export CLASSPATH="$(find core/target/ -type f -name '*.jar' | grep '\-package' | tr '\n' ':')"
```

used to produce a broken CLASSPATH and the `connect-standalone.sh` commands mentioned in the docs did not work.

### DIFF
--- a/core/src/main/assembly/package.xml
+++ b/core/src/main/assembly/package.xml
@@ -29,8 +29,8 @@
   <includeBaseDirectory>false</includeBaseDirectory>
   <fileSets>
     <fileSet>
-      <directory>${project.basedir}</directory>
-      <outputDirectory>share/doc/${project.name}/</outputDirectory>
+      <directory>${project.basedir}/..</directory>
+      <outputDirectory>share/doc/${project.artifactId}/</outputDirectory>
       <includes>
         <include>README*</include>
         <include>LICENSE*</include>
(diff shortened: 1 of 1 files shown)


orpiske:
> You're the first person reporting this. Building the project with mvn clean package and running with connect-standalone.sh always worked for everybody here.

I was puzzled by this as well, but I took a look at the jars and it seems some recent change is indeed causing the build to generate jars with strange names. 

```$ find core/target/ -type f -name '*.jar'
core/target/camel-kafka-connector-0.0.1-SNAPSHOT-package/share/java/Camel-Kafka-Connector :: Core/jackson-dataformat-cbor-2.10.1.jar
core/target/camel-kafka-connector-0.0.1-SNAPSHOT-package/share/java/Camel-Kafka-Connector :: ...```

oscerd:
I tested the stuff locally and PR seems ok. Lets wait for the build and merge.

ppalaga:
The issue is a consequence of this change that happened on 2019-12-13 https://github.com/apache/camel-kafka-connector/commit/08749ea487f6ff80ab6fe5727e7284b19921b946#diff-357e4854869b2e21c38b1b437f11095aR33

Issue #15, opened 2019-12-09
Find a way to manage dependencies


oscerd:
In order to support camel components, the proper dependencies must be in the classpath. Decide a way to do it, there are some alternatives:

- trivial option: put all the supported camel components as project dependencies
- generate a kafka connector for each component like more or less is done for camel spring boot starters

comment 1 by @oscerd: Maybe the second one and then release a bom with all the kafka-connector we have.

comment 2 by @omarsmak: @valdar @oscerd question wondering in my head, I think there are components IMHO don't make sense to be contained as Kafka Connect connector (e.g: activemq, osgi related ..etc), how do we plan to tackle these?

valdar:
Thanks to https://github.com/apache/camel-kafka-connector/pull/102 we have in place the structure and dependency generation of a Kafka connector for each component.

The missing bits, apart from polishing and inevitable issues, is to generate the 4 classes needed for each connector to configure and work properly.

PR #85, opened 2020-02-07
First step for issue #15

valdar:


tooling/camel-kafka-connector-generator-maven-plugin/src/main/java/org/apache/camel/springboot/maven/PrepareCatalogSpringBootMojo.java
```diff
@@ -0,0 +1,187 @@
+/*
```
oscerd: Is this Spring Boot stuff needed here?

core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java
```diff
@@ -42,6 +44,9 @@
 public class CamelSourceTask extends SourceTask {
```
omarsmak: Do you think it makes sense if we change the base classes `CamelSinkConnector`, `CamelSourceConnector`, `CamelSinkTask` and `CamelSourceTask` to abstract with abstract methods that need to be overriden? IMHO it makes it easier to follow that need to be overriden

connectors/camel-sjms2-kafka-connector/src/main/java/org/apache/camel/kafkaconnector/sjms2/CamelSjms2SourceConnectorConfig.java
```diff
@@ -0,0 +1,237 @@
...
+
+    public CamelSjms2SourceConnectorConfig(Map<String, String> parsedConfig) {
+        this(conf(), parsedConfig);
+    }
+
+    public static ConfigDef conf() {
+        ConfigDef conf = new ConfigDef(CamelSourceConnectorConfig.conf());
+
+        return conf.define(CAMEL_SOURCE_BRIDGE_ERROR_HANDLER_CONSUMER_CONF, ConfigDef.Type.BOOLEAN, CAMEL_SOURCE_BRIDGE_ERROR_HANDLER_CONSUMER_DEFAULT, ConfigDef.Importance.HIGH, CAMEL_SOURCE_BRIDGE_ERROR_HANDLER_CONSUMER_DOC)
+                .define(CAMEL_SOURCE_SJMS2_CONNECTION_FACTORY_CONF, ConfigDef.Type.STRING, CAMEL_SOURCE_SJMS2_CONNECTION_FACTORY_DEFAULT, ConfigDef.Importance.HIGH, CAMEL_SOURCE_SJMS2_CONNECTION_FACTORY_DOC)
```
omarsmak: How the importance being determined here? I guess it depends if the property is required?

valdar:
So @omarsmak @orpiske @oscerd I need your input and view on this (and form whoever likes to participate in the conversation).

As you can see there are 3 main parts to this PR:
- changes to `core/`: these are to enable pre-configuration of a connector by subclassing `Camel{Sink,Source}Connector` and `Camel{Sink,Source}Task` and providing the initial configuration through an override of `getDefaultConfig()`. Everything is still configurable through the property file supplied when you run the connector since those properties have precedence. All will still work with the old `camel.{sink,source}.url` config since if this one is present all the properties `camel.endpoint` `camel.path` and `camel.{sink,source}.component` will be ignored. If `camel.{sink,source}.url` is null though, the url is assembled by combining `camel.{sink,source}.component` and `camel.endpoint` `camel.path` accordingly.
- in `connectors/camel-sjms2-kafka-connector` there is the sjms2 connector done by hand has it should be generated.
- in `tooling/` I have pre imported the maven plugins from camel-spring-boot project, but those are not functional, will serve as a base for the prosecution of my work: the actual generations.

That said here is my question: if you look at `CamelSjms2SinkConnectorConfig` and `CamelSjms2SourceConnectorConfig` in `connectors/camel-sjms2-kafka-connector` you see that there are all those kafka connector properties definitions that can be generated parsing the `camel-catalog`; has that much value-added? Those properties are never used in the code, they are there only for documentation resons, would be better to just generate good, tweaked for connectors, documentation starting from the catalog?

tooling/camel-kafka-connector-generator-maven-plugin/src/main/java/org/apache/camel/springboot/maven/PrepareCatalogSpringBootMojo.java
valdar: See my last comment.

core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java
valdar: Once ve finished the whole autogeneration could be an option, the fact is, those are usable as they are, if you know what you are doing with the properties :)

connectors/camel-sjms2-kafka-connector/src/main/java/org/apache/camel/kafkaconnector/sjms2/CamelSjms2SourceConnectorConfig.java
valdar: Good observation, it wasn't determined. A criterion is good as any other I guess.

core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java
omarsmak: Yeah we can iterate over it once the autogen is finished :). Well, it is true, but will be nice to keep it obvious that it may need some implementation ;)

oscerd:
> So @omarsmak @orpiske @oscerd I need your input and view on this (and form whoever likes to participate in the conversation).
> 
> As you can see there are 3 main parts to this PR:
> 
> * changes to `core/`: these are to enable pre-configuration of a connector by subclassing `Camel{Sink,Source}Connector` and `Camel{Sink,Source}Task` and providing the initial configuration through an override of `getDefaultConfig()`. Everything is still configurable through the property file supplied when you run the connector since those properties have precedence. All will still work with the old `camel.{sink,source}.url` config since if this one is present all the properties `camel.endpoint` `camel.path` and `camel.{sink,source}.component` will be ignored. If `camel.{sink,source}.url` is null though, the url is assembled by combining `camel.{sink,source}.component` and `camel.endpoint` `camel.path` accordingly.
> * in `connectors/camel-sjms2-kafka-connector` there is the sjms2 connector done by hand has it should be generated.
> * in `tooling/` I have pre imported the maven plugins from camel-spring-boot project, but those are not functional, will serve as a base for the prosecution of my work: the actual generations.
> 
> That said here is my question: if you look at `CamelSjms2SinkConnectorConfig` and `CamelSjms2SourceConnectorConfig` in `connectors/camel-sjms2-kafka-connector` you see that there are all those kafka connector properties definitions that can be generated parsing the `camel-catalog`; has that much value-added? Those properties are never used in the code, they are there only for documentation resons, would be better to just generate good, tweaked for connectors, documentation starting from the catalog?

If this options are usable in a configuration file and thinking about the fact we still have to generate documentation for the website, I think it makes sense to have these properties available and documented. We can also try to parse the kafka config classes and generate the single connector documentation from them. WDYT?

omarsmak:
@valdar I think these configs could be used if you plan to start the these connectors programetely, as it makes it easier to just do something like `Properties.put(CamelSjms2SinkConnectorConfig.CAMEL_SINK_SJMS2_CONNECTION_RESOURCE_CONF, "test")`. If it is not too much of hassle, I'd prefer to keep them though

omarsmak:
@oscerd yeah we can use it `confDef` to generate the documentation, if not mistaken, Debezium uses the `confDef` to generate the documentation for the website

oscerd:
We can merge this once you fix the findings from @omarsmak and start from this base. @zregvart can you help on setting up the needed bits for publishing on the website? A simple page for the moment would be good, since we need to setting up the same for Spring boot support (camel-spring-boot repository) it would be good if you could help here or after the spring boot stuff. If you have time obviously.

connectors/camel-sjms2-kafka-connector/src/main/java/org/apache/camel/kafkaconnector/sjms2/CamelSjms2SourceConnectorConfig.java
omarsmak: No problems. We can iterate over it later since we will have this base. Thanks again for this !

orpiske:
> So @omarsmak @orpiske @oscerd I need your input and view on this (and form whoever likes to participate in the conversation).
> 
> As you can see there are 3 main parts to this PR:
> 
>     * changes to `core/`: these are to enable pre-configuration of a connector by subclassing `Camel{Sink,Source}Connector` and `Camel{Sink,Source}Task` and providing the initial configuration through an override of `getDefaultConfig()`. Everything is still configurable through the property file supplied when you run the connector since those properties have precedence. All will still work with the old `camel.{sink,source}.url` config since if this one is present all the properties `camel.endpoint` `camel.path` and `camel.{sink,source}.component` will be ignored. If `camel.{sink,source}.url` is null though, the url is assembled by combining `camel.{sink,source}.component` and `camel.endpoint` `camel.path` accordingly.
> 
>     * in `connectors/camel-sjms2-kafka-connector` there is the sjms2 connector done by hand has it should be generated.
> 
>     * in `tooling/` I have pre imported the maven plugins from camel-spring-boot project, but those are not functional, will serve as a base for the prosecution of my work: the actual generations.
> 
> 
> That said here is my question: if you look at `CamelSjms2SinkConnectorConfig` and `CamelSjms2SourceConnectorConfig` in `connectors/camel-sjms2-kafka-connector` you see that there are all those kafka connector properties definitions that can be generated parsing the `camel-catalog`; has that much value-added? Those properties are never used in the code, they are there only for documentation resons, would be better to just generate good, tweaked for connectors, documentation starting from the catalog?

This looks very good. My only comment here is regarding the properties. I think that even if they are not used in the code, there may be some value on them if they can be used for documentation properties. Even if they are not directly used, folks reading the code may still find some value in them. 

Everything else looks fine to me and I'm looking forward the changes.

valdar:
So to summarize  3 votes towards trying to generate the Kafka Connect `propertyDef` and on top of that generating the documentation (directly from them or from the catalog is to be understood).

I fear we might integrate the generation of the documentation from `propertyDef`, if you have noticed, complex Object properties are defined as STRING in the `propertyDef` since there is no other way that I have found.

@omarsmak since you mentioned it, can you have a look at what Debezium dose to generate the properties from `propertyDef`? Maybe we can open a separate issue for that?

omarsmak:
@valdar sure I can take a look at it on Monday and I will get back with the findings

omarsmak:
@valdar sorry I was wrong. Apparently DBZ connector properties in the docs is a manual process

PR #99, opened 2020-02-28
Bump to Camel 3.1.0

oscerd:


tests/pom.xml
```diff
@@ -174,6 +174,36 @@
             <groupId>org.apache.qpid</groupId>
             <artifactId>qpid-jms-client</artifactId>
             <scope>test</scope>
+            <exclusions>
```
valdar: Dose a `qpid-jms-client` exists with the "right" netty dependencies so, instead of adding exclusions, we can just bump `qpid-jms-client` to that version?

oscerd:
Moved to 0.49.0

PR #102, opened 2020-03-08
Auto gen step 2: generated connectors structure and dependencies.

valdar:


orpiske:
This is looking great.

I think we could add `camel.sink.path.destinationName` and `camel.source.path.destinationName` to their respective configuration classes so that it easier to refer to them on the tests (and for documentation as well). 

What do you think @valdar ?

valdar:
I don't fully get what you mean @orpiske :/

orpiske:
> I don't fully get what you mean @orpiske :/

No worries. My comment was not very clear. I was thinking of about some connector specific configuration that we had. 

For example, I was thinking of this setting we used to set the [JMS destination address](https://github.com/apache/camel-kafka-connector/blob/master/connectors/camel-sjms2-kafka-connector/src/main/java/org/apache/camel/kafkaconnector/sjms2/CamelSjms2SinkConnectorConfig.java#L152).   

I was just wondering/suggesting if it would OK to have those on the [CamelSinkConnectorConfig.java](https://github.com/apache/camel-kafka-connector/blob/master/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkConnectorConfig.java).

valdar:
It should be already there https://github.com/apache/camel-kafka-connector/blob/master/connectors/camel-sjms2-kafka-connector/src/main/java/org/apache/camel/kafkaconnector/sjms2/CamelSjms2SinkConnectorConfig.java#L152 or am I musunderstanding you again? :'(

orpiske:
> It should be already there https://github.com/apache/camel-kafka-connector/blob/master/connectors/camel-sjms2-kafka-connector/src/main/java/org/apache/camel/kafkaconnector/sjms2/CamelSjms2SinkConnectorConfig.java#L152 or am I musunderstanding you again? :'(

You are correct. I did have a failure here because that was missing, but I guess it was some problem on my end.

PR #107, opened 2020-03-18
Added support for remote AWS testing

orpiske:
Includes:
- Added remote testing support for AWS SNS sink
- Added remote testing support for AWS SQS source/sink
- Added remote testing support for AWS S3 source
- Added remote testing support for AWS Kinesis source
- Updated documentation for remote testing

PR #108, opened 2020-03-19
Avoid building all the code for sourcecheck verification

orpiske:


orpiske:
This avoids building and installing all of code twice and prevent the disk space problem that happened on PR #107

orpiske:
Also cuts down check time by +/- half of the time.

PR #111, opened 2020-03-23
Pointing to ASF Snapshot and use the apache Parent

oscerd:


valdar:
merged with check errors since was a github action issue.

PR #138, opened 2020-04-22
Test code fixes for the auto-generated connectors

orpiske:
This fixes a couple of errors in tests and adjust the code to work with the new
auto-generated connectors. It fixes most of the DelegatingClassLoader errors 
although one still remain in one scenario: ElasticSearch Rest with Java 8.

### DIFF
--- a/tests/src/test/java/org/apache/camel/kafkaconnector/PluginPathHelper.java
+++ b/tests/src/test/java/org/apache/camel/kafkaconnector/PluginPathHelper.java
@@ -79,7 +79,7 @@ public final class PluginPathHelper {
     }
 
     private static List<String> findPlugins() {
-        return findPlugins("core", "connectors/camel-sjms2-kafka-connector");
+        return findPlugins("core");
     }
 
     /*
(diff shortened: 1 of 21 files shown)


tests/src/test/java/org/apache/camel/kafkaconnector/PluginPathHelper.java
```diff
@@ -31,6 +31,14 @@
 public final class PluginPathHelper {
     private static final Logger LOG = LoggerFactory.getLogger(PluginPathHelper.class);
 
+    private static final String[] MODULES = {
+            "core", "connectors/camel-sjms2-kafka-connector", "connectors/camel-cql-kafka-connector", "" +
```
orpiske, line 35: Broken formatting. I'll fix.

oscerd:
@orpiske can you fix the codestyle?

orpiske:
> @orpiske can you fix the codestyle?

Done! It should be fine now.

oscerd:
Thanks!

orpiske:
@oscerd @valdar The code is failing because it is looking for the jars and we don't generate them here. I'm going to go back to the code and handle this scenario.

oscerd:
LGTM

oscerd:
@valdar is it ok for you?

orpiske:
BTW: to remove the ElasticSearch error, we need to find a way to stop needing [this class](https://github.com/apache/camel-kafka-connector/blob/master/tests/src/test/java/org/apache/camel/kafkaconnector/sink/elasticsearch/transforms/ConnectRecordValueToMapTransformer.java). It does the transformation from the key/pair data format used in ElasticSearch to the record format used by the Kafka message. If we have a way to do this conversion automatically, then we can remove this class and [the dependency](https://github.com/apache/camel-kafka-connector/blob/master/tests/pom.xml#L49) on the `camel-elasticsearch-rest-kafka-connector`. This should do the trick of removing the last DelegationClassLoader error on the logs.

PR #174, opened 2020-04-29
Adds a manual test for the Camel Slack sink connector

orpiske:
Due to being manual, the test is disabled by default. The PR also includes an example for the Slack sink connector


orpiske:
> We could also add this in https://github.com/apache/camel-kafka-connector-examples

Thanks. I'll send a PR there as well.

orpiske:
`[ERROR] /home/runner/work/camel-kafka-connector/camel-kafka-connector/tests/src/test/java/org/apache/camel/kafkaconnector/sink/slack/CamelSinkSlackITCase.java:20:8: Unused import - java.util.concurrent.CountDownLatch. [UnusedImports]` 

Fixing it. Strange that it did not catch here.

PR #193, opened 2020-05-06
Add contentLogLevel task property

fvaleri:
We should not log record's payload by default, as it may contain sensitive information. Anyway it is useful to have a configuration option to enable/disable it for debugging purpose.

Proposing new boolean config named `camel.source.logPayload`.

### DIFF
--- a/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkConnectorConfig.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkConnectorConfig.java
@@ -18,6 +18,7 @@ package org.apache.camel.kafkaconnector;
 
 import java.util.Map;
 
+import org.apache.camel.LoggingLevel;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
@@ -42,10 +43,15 @@ public class CamelSinkConnectorConfig extends AbstractConfig {
     public static final String TOPIC_CONF = "topics";
     public static final String TOPIC_DOC = "A list of topics to use as input for this connector";
 
+    public static final String CAMEL_SINK_CONTENT_LOG_LEVEL_DEFAULT = LoggingLevel.OFF.toString();
+    public static final String CAMEL_SINK_CONTENT_LOG_LEVEL_CONF = "camel.sink.contentLogLevel";
+    public static final String CAMEL_SINK_CONTENT_LOG_LEVEL_DOC = "Log level for the record's content (default: " + CAMEL_SINK_CONTENT_LOG_LEVEL_DEFAULT + "). Valid values: TRACE, DEBUG, INFO, WARN, ERROR, OFF.";
+
     private static final ConfigDef CONFIG_DEF = new ConfigDef()
         .define(CAMEL_SINK_URL_CONF, Type.STRING, CAMEL_SINK_URL_DEFAULT, Importance.HIGH, CAMEL_SINK_URL_DOC)
         .define(CAMEL_SINK_MARSHAL_CONF, Type.STRING, CAMEL_SINK_MARSHAL_DEFAULT, Importance.HIGH, CAMEL_SINK_MARSHAL_DOC)
-        .define(CAMEL_SINK_COMPONENT_CONF, Type.STRING, CAMEL_SINK_COMPONENT_DEFAULT, Importance.HIGH, CAMEL_SINK_COMPONENT_DOC);
+        .define(CAMEL_SINK_COMPONENT_CONF, Type.STRING, CAMEL_SINK_COMPONENT_DEFAULT, Importance.HIGH, CAMEL_SINK_COMPONENT_DOC)
+        .define(CAMEL_SINK_CONTENT_LOG_LEVEL_CONF, Type.STRING, CAMEL_SINK_CONTENT_LOG_LEVEL_DEFAULT, Importance.HIGH, CAMEL_SINK_CONTENT_LOG_LEVEL_DOC);
 
     public CamelSinkConnectorConfig(ConfigDef config, Map<String, String> parsedConfig) {
         super(config, parsedConfig);
--- a/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
@@ -110,6 +110,7 @@ public class CamelSinkTask extends SinkTask {
     @Override
     public void put(Collection<SinkRecord> sinkRecords) {
         for (SinkRecord record : sinkRecords) {
+            TaskHelper.logRecordContent(LOG, record, config);
             Map<String, Object> headers = new HashMap<String, Object>();
             Exchange exchange = new DefaultExchange(producer.getCamelContext());
             headers.put(KAFKA_RECORD_KEY_HEADER, record.key());
@@ -123,7 +124,8 @@ public class CamelSinkTask extends SinkTask {
             }
             exchange.getMessage().setHeaders(headers);
             exchange.getMessage().setBody(record.value());
-            LOG.debug("Sending {} to {}", exchange, LOCAL_URL);
+
+            LOG.debug("Sending exchange {} to {}", exchange.getExchangeId(), LOCAL_URL);
             producer.send(LOCAL_URL, exchange);
         }
     }
--- a/core/src/main/java/org/apache/camel/kafkaconnector/utils/TaskHelper.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/utils/TaskHelper.java
@@ -22,7 +22,14 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.camel.LoggingLevel;
 import org.apache.camel.catalog.RuntimeCamelCatalog;
+import org.apache.camel.kafkaconnector.CamelSinkConnectorConfig;
+import org.apache.camel.kafkaconnector.CamelSourceConnectorConfig;
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.connect.connector.ConnectRecord;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.slf4j.Logger;
 
 public final class TaskHelper {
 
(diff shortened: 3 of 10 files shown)


davsclaus:
It appears the existing logging is at DEBUG logging level. This new option dont mention this fact so I would assume end users think its logged at INFO logging or always. 

Also it may not only be message body that has sensitive data. Headers can contain autorization tokens, usernames etc. So its probably better to have this option for all of it or nothing.

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
```diff
@@ -179,7 +182,7 @@ public void testSourcePollingTimeout() throws InterruptedException {
         CamelSourceTask camelSourceTask = new CamelSourceTask();
         camelSourceTask.start(props);
 
-        long sleepTime = 30L;
+        long sleepTime = 11L;
```
oscerd, line 185: Is there a specific reason?

fvaleri:
> It appears the existing logging is at DEBUG logging level. This new option dont mention this fact so I would assume end users think its logged at INFO logging or always.

Yes now is at DEBUG level, which is often set when debugging issues. The proposed option logs at INFO level only if the user explicitly set `camel.source.logPayload=true`, so it's somewhat independent from log level (unless you set it to OFF). 

> Also it may not only be message body that has sensitive data. Headers can contain autorization tokens, usernames etc. So its probably better to have this option for all of it or nothing.

It's actually including all of this, maybe the option name is not clear enough. I'm open to suggestions for better wording.

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
fvaleri: While running tests, I found `30L` to fail sporadically.

davsclaus:
You can also have the option about the logging level, we do this in Camel, then you can set it to DEBUG, INFO, OFF etc

fvaleri:
> You can also have the option about the logging level, we do this in Camel, then you can set it to DEBUG, INFO, OFF etc

Good idea. Will do that.

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
valdar: If that is the case I think it should be addressed in a separate PR after opening an issue.

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
```diff
@@ -122,7 +122,14 @@ public void put(Collection<SinkRecord> sinkRecords) {
             }
             exchange.getMessage().setHeaders(headers);
             exchange.getMessage().setBody(record.value());
-            LOG.debug("Sending {} to {}", exchange, LOCAL_URL);
+
+            if (LOG.isDebugEnabled()) {
```
valdar, line 126: I seem to remember that in modern logging frameworks this is done already by using the logging placeholders syntax (i.e.`LOG.debug("... {}", yourValue);`) so it should be useless and just pollute the code, but I might be wrong.

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
```diff
@@ -122,7 +122,14 @@ public void put(Collection<SinkRecord> sinkRecords) {
...
+                LOG.debug("Sending exchange {} to {}", exchange.getExchangeId(), LOCAL_URL);
+            }
+            if (config.getBoolean(CamelSinkConnectorConfig.CAMEL_SINK_LOG_PAYLOAD_CONF)) {
+                LOG.info(TaskHelper.getExchangePayload(exchange));
```
valdar, line 130: As @davsclaus pointed out in a comment, sensitive data can be in other places than the body and is not clear from the option name that, if enabled, you log at info level (I would have left it to debug and harmonized somehow with the previuse logging statement in order not to duplicate logging entries).

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkConnectorConfig.java
```diff
@@ -38,10 +38,16 @@
     public static final String CAMEL_SINK_URL_DOC = "The camel url to configure the destination. If this is set " + CAMEL_SINK_COMPONENT_CONF
             + " and all the properties starting with " + CamelSinkTask.getCamelSinkEndpointConfigPrefix() + ".<" + CAMEL_SINK_COMPONENT_CONF + " value> are ignored.";
 
+    // do not log record's payload by default, as it may contain sensitive information
+    public static final Boolean CAMEL_SINK_LOG_PAYLOAD_DEFAULT = false;
+    public static final String CAMEL_SINK_LOG_PAYLOAD_CONF = "camel.source.logPayload";
```
valdar, line 43: This I think should be `"camel.sink.logPayload"` ?

oscerd:
For the Thread.sleep I'm already opening a PR using Awaitility

davsclaus:
btw for logging you can use CamelLogger that can do the INFO, DEBUG, OFF etc
https://github.com/apache/camel/blob/master/core/camel-api/src/main/java/org/apache/camel/spi/CamelLogger.java

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
fvaleri: Yep, you are right. It only makes sense if there is string concatenation (to avoid multiple String objects being created when the level is disabled), which is not the case here. Will revert it.

fvaleri:
> For the Thread.sleep I'm already opening a PR using Awaitility
Thanks

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkConnectorConfig.java
fvaleri: Thanks

fvaleri:
Added the log option at the record level. Build is failing because of that flaky test that @oscerd is fixing with another PR.

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
fvaleri: Reverted

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
fvaleri: Now log is on ConnectRecord, rather than Camel Exchange.

valdar:
@fvaleri can you please rebase this PR on `master` and `push -f` on the PR so a Github Action run is triggered afresh and we can see if the flakey tests are fixed?

fvaleri:
@valdar done, but I still see random build failures:
```[ERROR] Tests run: 9, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4.421 s <<< FAILURE! - in org.apache.camel.kafkaconnector.CamelSourceTaskTest
[ERROR] org.apache.camel.kafkaconnector.CamelSourceTaskTest.testUrlPrecedenceOnComponentProperty  Time elapsed: 0.026 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <2> but was: <1>
	at org.apache.camel.kafkaconnector.CamelSourceTaskTest.testUrlPrecedenceOnComponentProperty(CamelSourceTaskTest.java:230)```

**Comment by valdar** [comment]
Yep, I will look into in, prepare to more rebases :D

**Comment by fvaleri** [comment]
Yeah no problem, I'm using the following simple script to test and in addition to `testUrlPrecedenceOnComponentProperty` there is also `testSourcePolling` that remains flaky. We must get rid of all those sleeps, I will also have a look later.
```
#!/bin/sh
for (( i = 1; ; i++ ))
do
  echo "Attempt $i"
  mvn test -pl core -am
  exitcode=$?
  if [ $exitcode -ne 0 ]
  then
    echo "Error at attempt $i"
    exit
  fi
done
```

**Comment by valdar** [comment]
I fixed `testUrlPrecedenceOnComponentProperty` since the purpose of that test wasn't to count messages in a given time frame. Regarding `testSourcePolling` it is a little bit more complicated since the purpose of the test is to ensure that ~2 messages are received in the given unit of time... we can be a little bit more permissive there, by increasing the number of messages and account for a +/-1 error... I am open to suggestions though.

**Comment by valdar** [comment]
Actually nevermind, I think I have found a way to test what I needed without relying on `Thread.sleep(...)`. PR will follow, once it is merged you can rebase; thanks in advance!

**Comment by fvaleri** [comment]
Sent my proposal on a dedicated PR, hope it's useful.

**Comment by valdar** [comment]
@fvaleri can you rebase this on master (again :disappointed: ) so we (finally :partying_face: ) get it merged?

**Comment by fvaleri** [comment]
> If we want to use the logging level instead of the boolean parameter, then I would need to create an utility class with a switch-case method to not hard-wire knowledge of what logging system is behind the slf4j facade. In that case we could set it to OFF by default.

Sure let me do it. I'm quoting myself here with a different approach, but I would need to explore it. What do you think? I can explore also this approach.

**Comment by valdar** [comment]
@fvaleri if you want to look into use the log level instead of a boolean you are welcome to do so; before writing your own utility class though, take a look at what @davsclaus suggested: https://github.com/apache/camel/blob/master/core/camel-api/src/main/java/org/apache/camel/spi/CamelLogger.java it might have all the pieces you need.

**Comment by fvaleri** [comment]
I think it's a better approach but I need to find some time to play with it. Bear with me, I will update this PR.

File: core/src/main/java/org/apache/camel/kafkaconnector/utils/TaskHelper.java [hunk 5]
```diff
@@ -93,4 +101,40 @@ private static Boolean stringStartWithOneOfPrefixes(String s, Set<String> prefix
...
+            // do not log record's content by default, as it may contain sensitive information
+            LoggingLevel level = LoggingLevel.OFF;
+            try {
+                final String key = (record instanceof SourceRecord)
+                    ? CamelSourceConnectorConfig.CAMEL_SOURCE_CONTENT_LOG_LEVEL_CONF
+                    : CamelSinkConnectorConfig.CAMEL_SINK_CONTENT_LOG_LEVEL_CONF;
+                level = LoggingLevel.valueOf(config.getString(key).trim().toUpperCase());
+            } catch (Throwable e) {
+            }
+            if (level != LoggingLevel.OFF && Arrays.asList(LoggingLevel.values()).contains(level)) {
```
valdar, line 116: is this `Arrays.asList(LoggingLevel.values()).contains(level)` check needed? or you can just log a warning in the default clasue of the switch case?

valdar: Can you also fix conflict with `master`?

fvaleri: Conflicts fixed.

> is this `Arrays.asList(LoggingLevel.values()).contains(level)` check needed? or you can just log a warning in the `default` clasue of the `switch case`?

Added a warning message in case of invalid level, thanks.

> In the end have you decided to not use `CamelLogger`? I am not implying is wrong, I am just asking.

We can use it, but, unless I'm missing something, it seems that it does not support string substitution, so we would be back to unnecessary object creation.

valdar: +1 it was my impression as well.

valdar:
Thanks for the contribution @fvaleri !

PR #201, opened 2020-05-09
Fix CamelSourceTask flaky tests

fvaleri:
This seems to be stable on my machine.


valdar:
@fvaleri thanks for the PR, but I preferred to get rid of `sleep(...)` as in https://github.com/apache/camel-kafka-connector/pull/204 .

PR #205, opened 2020-05-11
Improved CamelSourceTaskTest

fvaleri:
Merged last commit from @valdar with my previous refactoring. There is no sleep time and I have also removed timer `delay=0` option that adds about 30s delay or more. Now, on my machine, the whole suite runs consistently in about 8s vs 38s of previous version.

### DIFF
--- a/core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
+++ b/core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
@@ -19,7 +19,6 @@ package org.apache.camel.kafkaconnector;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Stream;
 
 import org.apache.camel.ProducerTemplate;
 import org.apache.kafka.connect.data.Schema;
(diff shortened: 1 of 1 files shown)


fvaleri:
Yes, let me fix `testSourcePollingTimeout` and push force.

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
```diff
@@ -32,270 +31,239 @@
...
+        final long size = 2;
         Map<String, String> props = new HashMap<>();
-        props.put("camel.source.url", "direct:start");
-        props.put("topics", "mytopic");
+        props.put(CamelSourceConnectorConfig.TOPIC_CONF, TOPIC_NAME);
+        props.put(CamelSourceConnectorConfig.CAMEL_SOURCE_URL_CONF, DIRECT_URI);
 
-        CamelSourceTask camelSourceTask = new CamelSourceTask();
-        camelSourceTask.start(props);
+        CamelSourceTask sourceTask = new CamelSourceTask();
```
oscerd, line 51: Is there a particular reason to change variable names?

fvaleri: It's part of the refactoring, we have only CamelSourceTask, so it's redundant.

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
```diff
@@ -32,270 +31,239 @@
...
+        assertNull(poll4.get(0).keySchema());
+        assertNull(poll4.get(0).value());
+        assertNull(poll4.get(0).valueSchema());
+
+        sourceTask.stop();
     }
 
     @Test
-    public void testSourceUrlPrecedenceOnComponentProperty() {
+    public void testUrlPrecedenceOnComponentProperty() {
```
valdar, line 192: I think this version is not testing what the old version was testing since `endpointProperty` and `pathChunk` are not legal properties of `direct` component.

valdar:
@fvaleri the amount of time spared is largely due to the fact that now 100 messages are sent instead of 999. I agree that at present 100 should do since in 1 millisecond is unlikely that all of them will be consumed even on a faster machine than ours (we don't know what future will hold though :) )

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
fvaleri: Fixed, thanks.

fvaleri:
> @fvaleri the amount of time spared is largely due to the fact that now 100 messages are sent instead of 999. I agree that at present 100 should do since in 1 millisecond is unlikely that all of them will be consumed even on a faster machine than ours (we don't know what future will hold though :) )

Based on my tests, the difference between 100 and 999 is negligible as it is a tight loop (so I will revert back to 999). What really slow down is the use of timer's delay option, even if we set 0 or -1, there is about 30s delay. Then, using a classic `fori` instead of Stream's `foreach` to send batch test records is also faster.

core/src/test/java/org/apache/camel/kafkaconnector/CamelSourceTaskTest.java
valdar: Thanks!

valdar:
@fvaleri ah that is weird, surely can be explained digging into camel timer component, but I agree that we can implement the same tests avoiding `timer delay` option.

valdar:
@fvaleri thanks a lot for the PR and the patience. I am merging this with faling checks because they seem due to github actions problems.

fvaleri:
@valdar no problem at all, I'm happy to contribute. Thanks.

PR #218, opened 2020-05-18
Remove core from test plugin path

orpiske:
Core is not needed on the plugin path since test code now uses the connectors' path


oscerd:
Leave this parked until we finish the 0.2.0 release.

PR #243, opened 2020-05-25
Use camel-depencencies to import versions of common libraries like Ja…

### LINKED ISSUE #241: Use camel-depencencies to import versions of common libraries like Jakson, qpid, log4j etc.
Since right now we align these types of dependencies manually, probably is worth exploring if we can minimize that using `camel-dependencies` bom.

oscerd:
…kson, qpid, log4j etc and fixed camel-syslog-connector assembly

Fixes #241

### DIFF
--- a/connectors/camel-syslog-kafka-connector/src/main/assembly/package.xml
+++ b/connectors/camel-syslog-kafka-connector/src/main/assembly/package.xml
@@ -37,8 +37,8 @@
       </includes>
     </fileSet>
     <fileSet>
-      <directory><#noparse>${project.basedir}/src/main/docs</#noparse></directory>
-      <outputDirectory><#noparse>docs/</#noparse></outputDirectory>
+      <directory>${project.basedir}/src/main/docs</directory>
+      <outputDirectory>docs/</outputDirectory>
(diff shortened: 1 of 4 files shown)


connectors/camel-syslog-kafka-connector/src/main/assembly/package.xml
```diff
@@ -37,8 +37,8 @@
       </includes>
     </fileSet>
     <fileSet>
-      <directory><#noparse>${project.basedir}/src/main/docs</#noparse></directory>
-      <outputDirectory><#noparse>docs/</#noparse></outputDirectory>
```
oscerd, line 41: I think this was wrong @valdar

parent/pom.xml
```diff
@@ -27,18 +27,11 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <activemq.version>5.15.12</activemq.version>
         <kafka.version>2.4.1</kafka.version>
```
oscerd, line 30: This is still here, because camel 3.3.0 is on 2.5.0

PR #247, opened 2020-05-27
Adds integration tests for Salesforce

orpiske:
This adds some simple integration tests for Salesforce. Not all functionality
is covered, but I think it provides an initial infrastructure for future tests.

Currently covered:
- Source: basic data consumption tests + change data capture tests
- Sink: createSObject operation

The tests do not run by default. Because the Salesforce DTOs are required and
also some setup is required on Salesforce, the tests need to be enabled using
a maven profile (-Psalesforce) along with the required properties for the test
execution (all documented on the test classes).

Once this initial setup is done, the tests can run automatically.

### DIFF
--- a/docs/modules/ROOT/pages/contributing.adoc
+++ b/docs/modules/ROOT/pages/contributing.adoc
@@ -97,4 +97,68 @@ mvn -Dkafka.bootstrap.servers=host1:port -Dkafka.instance.type=remote -DskipInte
 
 It's possible to use a properties file to set these properties. To do so use `-Dtest.properties=/path/to/file.properties`.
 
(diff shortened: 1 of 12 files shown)


valdar:
Ah, thanks a lot @orpiske! Very nice and needed contribution.

Could it be possible to add the documented steps on a readme?

orpiske:
> Ah, thanks a lot @orpiske! Very nice and needed contribution.

Thanks for the review! I will include more details about how to run these tests :)

orpiske:
I added the details required to setup the Salesforce account, with some links to the documentation and the steps required to run the tests.

Does it look OK?

PR #254, opened 2020-06-01
In case of Sink we need to use toD instead of to

### LINKED ISSUE #252: In case of Sink we need to use toD instead of to
https://github.com/apache/camel-kafka-connector/blob/master/core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelMainSupport.java#L108

Actually changing to toD, make it possible to declare uriParam by using simple language.

oscerd:
Fixes #252


valdar:
It seems there are legitimate test failures related to Sink that might be related to the change in the commit.

oscerd:
I'll investigate

Issue #211, opened 2020-05-14
Improve integration test support when component dependencies are present


orpiske:
In a scenario where you are testing Camel component A and your test depends on Camel component B is necessary to include Camel component B as a provided dependency on the test pom. 

This makes testing more complicated, pollutes the pom and makes it harder to manage test dependencies.

orpiske:
Based on a discussion with @valdar

valdar:
It is mildly related to #191

valdar:
Easy and useful mitigation would be to have each connector tests under a submodule of the `tests` project, so one can better control the tests classpath by having only the dependencies needed by a specific connector.

orpiske:
> Easy and useful mitigation would be to have each connector tests under a submodule of the `tests` project, so one can better control the tests classpath by having only the dependencies needed by a specific connector.

+1. I think it would also make it easier to look at the tests logs. Right now, if we run all the tests at once, we end up with large log which is not always easy to read.

orpiske:
@valdar @oscerd I think we can close this since we split the tests in modules. What do you think?

PR #262, opened 2020-06-08
Split the tests in modules per service

orpiske:
This is a proposed change to solve the dependencies problems discussed on #211. 

This changes breaks down the test module into submodules and groups the tests per service or service group (ie.: AWS, SJMS2, ElasticSearch) etc. This way, we can reduce the number of dependencies in test time which would help us find problems such as missing dependencies on the connector.

As a result of this break-down, the logs are generated per module and the plugin.path configuration of the runtime loads only the connector under test. 

There is a small increase in test time, as the embedded Kafka and the Kafka Connect Runtime are recycled for every test.


orpiske:
Failing w/ a Java heap error. I am looking at it.

orpiske:
> Failing w/ a Java heap error. I am looking at it.

Hooray. The test have passed.

PR #272, opened 2020-06-11
Issue/260

valdar:


tests/itests-common/src/test/java/org/apache/camel/kafkaconnector/common/services/kafka/KafkaService.java
```diff
@@ -18,15 +18,13 @@
...
 import org.junit.jupiter.api.extension.AfterAllCallback;
-import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
-import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
 /**
  * Provides an interface for any type of Kafka service: remote instances, local container, etc
  */
-public interface KafkaService extends BeforeAllCallback, BeforeTestExecutionCallback, AfterAllCallback, AfterTestExecutionCallback {
```
orpiske, line 29: Out of curiosity: did you notice any problems w/ loading the plugins on the Connect runtime? One of the reasons I had to change this on #262 was because keeping the connect runtime instance up and just bouncing it after the whole module was done, leads to a leak on the runtime (which caused the first version of the PR to fail). Bouncing both solved the issue (at the expense of a longer test time).

valdar: Ah! Yep, that is what is happening in the GitHub action execution as well, thanks for the explanation and context about this!

valdar:
@oscerd thanks I will do that :/

PR #273, opened 2020-06-12
Throw ConnectException due to failed exchanges

omarsmak:
In case we have failed Camel exchanges, we throw  `ConnectException` in order to avoid the committing the offsets on failed messages and hence result of loss messages downstream. With this small modification, the task will directly shutdown upon failed exchanges and hence avoid any offsets being committed


valdar:
@omarsmak error handling is a theme in need of more love, so glad you contributed there!

Because it is of great importance as well can I ask you if you could add a test as well to this PR?

omarsmak:
Indeed @valdar, unit tests here are very important to ensure this always checked as such small modification has a significant semantics impact. I would definitely update this PR with unit tests. Will keep you posted!

omarsmak:
@valdar added unit tests

PR #295, opened 2020-06-23
Add tests for MongoDB

orpiske:


oscerd:
This was failing for hdfs. Now that test has been disabled.

PR #316, opened 2020-07-08
Added tests for JDBC sink

orpiske:


parent/pom.xml
```diff
@@ -301,6 +302,18 @@
...
+                <groupId>org.testcontainers</groupId>
+                <artifactId>mariadb</artifactId>
+                <version>${testcontainers-version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.mariadb.jdbc</groupId>
+                <artifactId>mariadb-java-client</artifactId>
+                <version>${version.mariadb}</version>
+                <scope>test</scope>
```
oscerd, line 315: mariadb-java-client is licensed LGPL, it's not compliant with ASF.

orpiske: Thanks @oscerd. I wasn't aware of that. I will adjust it :)

orpiske: @oscerd actually, I am thinking about refactoring this and using Postgres instead ... otherwise we'll have to set the test as manual. The Postgres one uses the BSD 2-clause. Is there any place I could check about the compatibility of this?

oscerd: Here https://www.apache.org/legal/resolved.html

BSD-2 should be ok

Issue #288, opened 2020-06-22
Enable testRetrieveFromCassandraWithCustomStrategy on CamelSourceCassandraITCase


orpiske:
The [testRetrieveFromCassandraWithCustomStrategy](
https://github.com/apache/camel-kafka-connector/blob/5090422c4faa6cfe2f506a5a53d14a875af5b654/tests/itests-cassandra/src/test/java/org/apache/camel/kafkaconnector/cassandra/source/CamelSourceCassandraITCase.java#L117) is currently disabled because of [CAMEL-15219](https://issues.apache.org/jira/browse/CAMEL-15219). This issue is now fixed on Camel and we can enable the test once/if we move to Camel 3.4.1 or greater.

orpiske:
Closing this one. It was enabled with #337.

PR #331, opened 2020-07-23
Re-enable a Cassandra test disabled due to CAMEL-15219

orpiske:
This test was disabled due to https://issues.apache.org/jira/browse/CAMEL-15219 which was fixed in Camel 3.4.2.


orpiske:
This should allow us to close #288

oscerd:
Seems to fail for checkstyle only

orpiske:
> Seems to fail for checkstyle only



> Seems to fail for checkstyle only

I forgot to remove the import for the disabled annotation. Updating and sending a new one in a bit.

orpiske:
@oscerd that's very strange. I did test w/ one of the 3.4.x snapshots when the issue was fixed on Camel and it did work. Of course, something is not right now. I'll cancel this PR and will dig into the issue.

orpiske:
Closed: needs investigation.

PR #336, opened 2020-07-26
Add CORE protocol support to SJMS2 connector

fvaleri:
I would be good to also support the CORE protocol with the SJMS2 connector.


docs/modules/ROOT/pages/connectors.adoc
```diff
@@ -71,6 +71,7 @@ Number of Camel Kafka connectors: 351
...
 | *camel-beanstalk-kafka-connector* | true | true | xref:connectors/camel-beanstalk-kafka-sink-connector.adoc[Sink Docs] | xref:connectors/camel-beanstalk-kafka-source-connector.adoc[Source Docs] | https://repo.maven.apache.org/maven2/org/apache/camel/kafkaconnector/camel-beanstalk-kafka-connector/0.3.0/camel-beanstalk-kafka-connector-0.3.0-package.zip[Download Zip] | https://repo.maven.apache.org/maven2/org/apache/camel/kafkaconnector/camel-beanstalk-kafka-connector/0.3.0/camel-beanstalk-kafka-connector-0.3.0-package.tar.gz[Download Tar.gz]
+| *camel-bonita-kafka-connector* | true | false | xref:connectors/camel-bonita-kafka-sink-connector.adoc[Sink Docs] |  | https://repo.maven.apache.org/maven2/org/apache/camel/kafkaconnector/camel-bonita-kafka-connector/0.3.0/camel-bonita-kafka-connector-0.3.0-package.zip[Download Zip] | https://repo.maven.apache.org/maven2/org/apache/camel/kafkaconnector/camel-bonita-kafka-connector/0.3.0/camel-bonita-kafka-connector-0.3.0-package.tar.gz[Download Tar.gz]
```
oscerd, line 74: Bonita has been removed

parent/pom.xml
```diff
@@ -287,6 +287,12 @@
                 <version>${activemq-version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
```
oscerd, line 290: If this is used only in the tests for sums it makes sense to add the dependency only in the specific test folder and not in parent

fvaleri:
Tested with [this example](https://github.com/fvaleri/cdc/tree/master/connect-cdc) changing the `camel.component.sjms2.connection-factory` to `#class:org.apache.activemq.ActiveMQConnectionFactory`.

connectors/camel-sjms2-kafka-connector/pom.xml
```diff
@@ -48,7 +48,10 @@
     <dependency>
       <groupId>org.apache.activemq</groupId>
       <artifactId>activemq-client</artifactId>
-      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.activemq</groupId>
+      <artifactId>artemis-jms-client</artifactId>
```
valdar, line 54: Is this client capable of all protocols that `activemq-client` support? In other words can we use just this dependences and get rid of `activemq-client`?

fvaleri: Nope, unfortunately only Artemis CORE. To support OpenWire, AMQP and CORE we need three distinct dependencies. Defaulting to OpenWire (activemq-client) is good because the connector works with both ActiveMQ and Artemis.

valdar: Got you, thanks!

Issue #178, opened 2020-04-30
Automate bootstrap for non-containerized deployment?


orpiske:
Looking at the [AWS-to-JMS](https://github.com/orpiske/camel-kafka-connector-examples/tree/master/aws-s3-to-jms) example it seems to me that the bootstrap steps are going to be similar most of the time. 

1. Download the connector package
2. (Optional) Download the dependencies
3. Create the configuration file
4. Put the files on the Kafka Connect runtime install directory. 

I know this is handled automatically that for containerized deployments such as Strimzi, so I was wondering if the project should have a facility for regular deployments. 

What do you think? IMHO, it could make even the documentation and example creation simpler, as the project can abstract a few repetitive things and reduce the entry barrier.

valdar:
In which form would you plan to ship that?

orpiske:
> It sounds a good idea to me! In which form would you plan to add that?

Thanks. 

I was thinking about a CLI utility, in a module within the project. The usage would look like this:

```camel-kafka-connector bootstrap --connector sjms2 --target-dir /path/to/kafka/connect```

It would have 2 additional arguments:

```
--download-host: to allow choosing the host from where to download the files
--override-property: to allow the user to override a setting
```

We already have the list of configurations in the connector config files ([example](https://github.com/apache/camel-kafka-connector/blob/master/connectors/camel-activemq-kafka-connector/src/main/java/org/apache/camel/kafkaconnector/activemq/CamelActivemqSinkConnectorConfig.java)).

It's not clear to me yet, how to handle generating the configuration. I think there are 2 options:

1. We could simply generate a configuration file with all the settings. 
2. We could generate a configuration file w/ only the required settings.  

Option 1 is significantly easier, although the configuration file would be much larger than our examples. Option 2 is more complicated and we would probably need to come up with a mechanism to highlight what is required and what is not. 

IMHO, I think we could stick w/ option 1.

lburgazzoli:
looks looks we may need a start.spring.io alike solution but for camel

valdar:
@orpiske For the configuration part actually I was planning to add an autogenerated example configuration listing only the required options (the camel catalog should provide that information). We can potentially convey that information using the Kafka connect config priority field (i.e. only the properties with priority = HIGH are the mandatory ones).

Probably we would need to specify the version as well? Or directly some GAV?

[...]

PR #343, opened 2020-07-30
Set default and doc config properties as private

fvaleri:
It's safer to not expose these kind of properties to prevent it from diverging from the documentation.


fvaleri:
I need to fix import order, but what to you think?

fvaleri:
I opened this for discussion too. I was thinking specifically about testing. In general it is a good idea to not expose what you don't need.

orpiske:
IMHO, we should keep them as public. One of the ideas we discussed (#178) could benefit from having access to these fields. Having access to the list of configurations and descriptions could, potentially, open the room for interfaces that validate and display those options in a way that is easier for the user to figure them out. 

Since they are ready-only already, it shouldn't be a problem as they are, IMHO.

fvaleri:
Hi @orpiske I see your point, but we have the `ConfigDef.getConfigValue` that we could use to expose these values when needed. In that case we would just need the map of `ConfigKey`s from the public method `ConfigDef.configKeys`.

orpiske:
> Hi @orpiske I see your point, but we have the `ConfigDef.getConfigValue` that we could use to expose these values when needed. In that case we would just need the map of `ConfigKey`s from the public method `ConfigDef.configKeys`.

Indeed, that could do the trick in this case!

oscerd:
I still have troubles to see the advantages, except good practice

fvaleri:
Hi @oscerd , there are no main advantages, so I'm ok to close it.

oscerd:
Wait, for @valdar and @orpiske too. I'm not the only here.if this is something they vote for, we'll merge it

fvaleri:
OK, I'm reopening it to wait for more feedback, but will fix the checkstyle issue only if there is consensus.

orpiske:
I agree that moving to private would follow generally accepted good practices (for lack of a better "term"). Howev er, I still feel -1 towards making it private. My feeling towards this comes from the fact that Kafka seems to also use [public visibility](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/ConnectorConfig.java#L98) on their code base.

Even though these are separate projects, I believe that maintaining them as `public` would be consistent to how it is done right now on the Kafka Connect code base. I believe that maintaining this consistency could simplify the job for developers looking to extend the Camel Kafka Connect connectors.

Therefore, my suggestion is -1.

fvaleri:
> I agree that moving to private would follow generally accepted good practices (for lack of a better "term"). However, I still feel -1 towards making it `private`. My feeling towards this comes from the fact that Kafka seems to also use [public visibility](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/ConnectorConfig.java#L98) on their code base.
> 
> Even though these are separate projects, I believe that maintaining them as `public` would be consistent to how it is done right now on the Kafka Connect code base. I believe that maintaining this consistency could simplify the job for developers looking to extend the Camel Kafka Connector connectors.
> 
> Therefore, my suggestion is -1.

Well, I see a mixed approach in that class. There are some `_DOC` and `_DISPALY` properties that are actually private, which is what I'm proposing here (make private every metadata associated with a given key). The `_CONF` properties would remain public, because they are actually needed outside of this class.

That said, I understand that this is more a matter of good practices than a real need, so I'm ok with whatever we choose.

orpiske:
> > I agree that moving to private would follow generally accepted good practices (for lack of a better "term"). However, I still feel -1 towards making it `private`. My feeling towards this comes from the fact that Kafka seems to also use [public visibility](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/ConnectorConfig.java#L98) on their code base.
> > Even though these are separate projects, I believe that maintaining them as `public` would be consistent to how it is done right now on the Kafka Connect code base. I believe that maintaining this consistency could simplify the job for developers looking to extend the Camel Kafka Connector connectors.
> > Therefore, my suggestion is -1.
> 
> Well, I see a mixed approach in that class. There are some `_DOC` and `_DISPALY` properties that are actually private, which is what I'm proposing here (make private every metadata associated with a given key). The `_CONF` properties would remain public, because they are actually needed outside of this class.
> 
> That said, I understand that this is more a matter of good practices than a real need, so I'm ok with whatever we choose.

Thanks for the clarification. I had initially thought that this would change the _CONF ones too. With your explanation I see that's not the case. So, given that it - with this change - it would be consistent with the way Kafka Connect code base is doing ... That's +1 from me.

valdar:
ah sorry to have missed this one, yeah I don't have a strong opinion in either direction, I am fine with the closing. (can always be revisited in the future).

PR #366, opened 2020-08-07
Building endpoint uris should for secret options automatic use RAW() …

davsclaus:
…syntax for the value so we wont have encoding issue for passwords or access tokens etc.

Here is a a bit rushed PR for a potential solution. When debugging I can see the endpoint is created with RAW() for those secret options.


oscerd:
Thanks! Let's wait for CI :-)

Issue #376, opened 2020-08-17
Unclean startup causes an NPE on shutdown


orpiske:
If the connector fails to start for any reason, it throws a NPE during shutdown because it tries to de-reference a null pointer. 

```
[2020-08-17 13:45:20,965] WARN Could not stop task (org.apache.kafka.connect.runtime.WorkerSinkTask:165)
org.apache.kafka.connect.errors.ConnectException: Failed to stop Camel context
        at org.apache.camel.kafkaconnector.CamelSinkTask.stop(CamelSinkTask.java:158)
        at org.apache.kafka.connect.runtime.WorkerSinkTask.close(WorkerSinkTask.java:163)
        at org.apache.kafka.connect.runtime.WorkerTask.doClose(WorkerTask.java:163)
        at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:190)
        at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:234)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NullPointerException
        at org.apache.camel.kafkaconnector.CamelSinkTask.stop(CamelSinkTask.java:156)
```

I'll send a PR for this.

orpiske:
Merged. Closing.

PR #377, opened 2020-08-17
Prevent NPEs on shutdown when critical errors occur

orpiske:


orpiske:
This is a proposed fix for #376

Issue #381, opened 2020-08-18
Discussion: troubleshooting documentation?


orpiske:
Do you think we should have a troubleshooting documentation/page with some of the common errors and problems that people face when they use the connectors? Some things I have in mind that we could add some notes about: 

- Installing the connectors in a directory structure that is incompatible with Kafka Connect
- Missing or incorrect converters
- Missing dependencies
- Known incompatibilities 
- Known limitations

oscerd:
+1

PR #383, opened 2020-08-18
Added a simple troubleshooting guide

orpiske:
A basic troubleshooting guide as proposed on #381


orpiske:
I am marking this one as a draft as there may be more stuff to add/adjust/etc and feedback too.

orpiske:
> I am marking this one as a draft as there may be more stuff to add/adjust/etc and feedback too.

Unmarked as draft. I think I added all initial doc bits I had in mind.

PR #391, opened 2020-08-25
Fix cassandra for camel 3.5

orpiske:
Adjusts the client code to use the newer Cassandra client version from
camel-cassandraql


oscerd:
Thanks a lot!

PR #396, opened 2020-08-26
Fix mongodb test

unsortedhashsets:
We need to pass variable to the mongodb.host in format mongodb://user:pass@domain otherway:
[ERROR]   CamelSinkMongoDBITCase.setUp:63 » IllegalArgument The connection string is inv...
[ERROR]   CamelSourceMongoDBITCase.setUp:66 » IllegalArgument The connection string is i...

But if we pass in that format we receive mongodb://mongodb://user:pass@domain:port and:
[ERROR] Failures: 
[ERROR]   CamelSinkMongoDBITCase.testBasicSendReceive:120->runTest:100 Condition not met within timeout 30000. The connector CamelMongoDBSinkConnector did not start within a reasonable time
[ERROR] Errors: 
[ERROR]   CamelSourceMongoDBITCase.testFindAll » Timeout testFindAll() timed out after 9...

With that fix - test will pass (tested with remote mongodb)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -65,8 +65,7 @@ for remote testing:
 ** hdfs.host
 ** hdfs.port
 * mongodb.instance.type
-** mongodb.host
-** mongodb.port
+** mongodb.url
 
 Additionally, a few manual tests can be enabled and executed with adequate configuration on the accounts and
 environments used by those services. This is very specific to the nature of each of those services, therefore
--- a/tests/itests-mongodb/src/test/java/org/apache/camel/kafkaconnector/mongodb/services/MongoDBLocalContainerService.java
+++ b/tests/itests-mongodb/src/test/java/org/apache/camel/kafkaconnector/mongodb/services/MongoDBLocalContainerService.java
@@ -41,18 +41,8 @@ public class MongoDBLocalContainerService implements MongoDBService {
     }
 
     @Override
-    public String getReplicaSetUrl() {
-        return container.getReplicaSetUrl();
-    }
-
-    @Override
-    public String getHost() {
-        return container.getContainerIpAddress();
-    }
-
-    @Override
-    public int getPort() {
-        return container.getMappedPort(DEFAULT_MONGODB_PORT);
+    public String getReplicaSetUrl() {    
+        return "mongodb://" + container.getContainerIpAddress() + ":" + container.getMappedPort(DEFAULT_MONGODB_PORT);
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


oscerd (changes requested):
@orpiske can you have a look? This will break the build I suppose

orpiske:
@oscerd sure thing, I will take a look :)

orpiske:
@unsortedhashsets thanks for the PR. I understand the problem you are trying to solve. Unfortunately, this way it breaks the build. The build test is failing on local mode because on local mode the service only provides the host and port. You need to adjust the patch a little bit in order to achieve the results you need in both local and remote test mode.

I would suggest replacing the `getHost` and `getPort` methods on the service [1] (and their local [2] and remote [3] implementations) with a single `getConnectionUrl` that provides the connection URL `mongodb://mongodb://user:pass@domain:port`. Therefore, instead of running the test like `-Dmongodb.host=mongodb://user:password@myhostname.com -Dmongodb.port=31244` you could pass it like this `-Dmongodb.url=mongodb://user:password@myhostname.com:31244`.

Of course, this is just a suggestion and there may be other ways to solve this issue, but I think this would do the trick. If you have questions, feel free to ask I can guide you. 

1. https://github.com/apache/camel-kafka-connector/blob/master/tests/itests-mongodb/src/test/java/org/apache/camel/kafkaconnector/mongodb/services/MongoDBService.java#L38
2. https://github.com/apache/camel-kafka-connector/blob/master/tests/itests-mongodb/src/test/java/org/apache/camel/kafkaconnector/mongodb/services/MongoDBLocalContainerService.java
3. https://github.com/apache/camel-kafka-connector/blob/master/tests/itests-mongodb/src/test/java/org/apache/camel/kafkaconnector/mongodb/services/RemoteMongoDBService.java

orpiske:
@unsortedhashsets I've just tested the changes and they are working well for local mode. 

c/c @oscerd

oscerd (approved):

unsortedhashsets:
Oh, sorry I have just finished remote test (success) and updated repo .

oscerd (approved):
LGTM, @orpiske can you retake a look?

orpiske:
> LGTM, @orpiske can you retake a look?

Sure thing! I have just run the test again, with the latest changes from @unsortedhashsets and it is working. IMHO, it's good to merge.

oscerd:
Thanks!

Issue #414, opened 2020-09-02
High CPU Usage with RabbitMQ Connector (polling interface implemented incorrectly?)


Thylossus:
Hi,

I'm using the [Camel Rabbit MQ Connector](https://camel.apache.org/camel-kafka-connector/latest/connectors/camel-rabbitmq-kafka-source-connector.html) as a source connector for my Kafka cluster. Kafka connect is running in a Docker container built with the following Dockerfile

```Dockerfile
FROM openjdk:8-jre-alpine

ARG VERSION_APACHE_KAFKA=2.2.1
ARG VERSION_SCALA=2.12
ARG VERSION_APACHE_CAMEL=0.4.0

RUN mkdir -p /kafka/connect/plugins && \
    mkdir -p /kafka/connect/app && \
    mkdir -p /kafka/connect/config && \
    wget https://archive.apache.org/dist/kafka/${VERSION_APACHE_KAFKA}/kafka_${VERSION_SCALA}-${VERSION_APACHE_KAFKA}.tgz -q -O kafka.tgz && \
    tar -xzf kafka.tgz -C /kafka/connect/app --strip 1 && \
    rm -f kafka.tgz && \
    wget https://repo1.maven.org/maven2/org/apache/camel/kafkaconnector/camel-rabbitmq-kafka-connector/${VERSION_APACHE_CAMEL}/camel-rabbitmq-kafka-connector-${VERSION_APACHE_CAMEL}-package.zip -q -O rabbitmq-connector.zip && \
    unzip -qq rabbitmq-connector.zip -d /kafka/connect/plugins && \
    rm -f rabbitmq-connector.zip && \
    apk add -q --no-cache bash

ENTRYPOINT [ "/kafka/connect/entrypoint.sh" ]
```

When running this configuration and adding a RabbitMQ connector with the following configuration

```json
{
  "name": "{{ config.connector_name }}",
  "config": {
    "connector.class": "org.apache.camel.kafkaconnector.rabbitmq.CamelRabbitmqSourceConnector",
    "topics": "{{ config.topic }}",
    "value.converter": "org.apache.kafka.connect.converters.ByteArrayConverter",
    "tasks.max": 1,
    "camel.component.rabbitmq.hostname": "{{ kafka_connect_rabbitmq_host }}",
    "camel.component.rabbitmq.portNumber": {{ kafka_connect_rabbitmq_port }},
[...]

PR #428, opened 2020-09-07
Proposed fix for issue #414

orpiske:


oscerd (approved):
LGTM, lets wait for @valdar

oscerd:
@orpiske can you please rebase so your branch will pick up the newly poiting to master gh actions? Thanks

orpiske:
> @orpiske can you please rebase so your branch will pick up the newly poiting to master gh actions? Thanks

Sure thing.

orpiske:
I think it should be ok now, @oscerd

oscerd:
Thanks, I think @valdar is reviewing.

valdar:
Hi @orpiske great work!

I was wondering which one of the contributions is predominant: better handling the remaining time or returning null if there are no results.

Do you mind, since you already have the performance tests at hand, doing a quick comparison towards just returning `null` if the result set is empty, like: https://github.com/valdar/camel-kafka-connector/commit/2edcfc079134ec2742cac8e5fa625170f4141373

core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java
```diff
@@ -119,53 +119,63 @@ public void start(Map<String, String> props) {
...
+        long remaining = remaining(startPollEpochMilli, maxPollDuration);
         long collectedRecords = 0L;
 
-        List<SourceRecord> records = new ArrayList<>();
-        while (collectedRecords < maxBatchPollSize && (Instant.now().toEpochMilli() - startPollEpochMilli) < maxPollDuration) {
-            Exchange exchange = consumer.receiveNoWait();
+        List<SourceRecord> records = null;
+        while (collectedRecords < maxBatchPollSize && remaining > 0) {
+            Exchange exchange = consumer.receive(remaining);
+            if (exchange == null) {
```
valdar, line 137: what is the point of this check? if the exchange is null, since you receive waiting for the whole remaining time it means you have finished the time so there is no point in updating it earlier and continue. 

I would rather just check if exchange is not `null` and in that case do all the work except updating remaining time, i.e. something like: 
```java
while(...) {
    Exchange exchange = consumer.receive(remaining);
    if(exchange != null) {
    //whole code here
    }
    remaining = remaining(startPollEpochMilli, maxPollDuration);
}
...
```

valdar: I might have completely overlooked something here not understanding your motivations, feel free to disagree and point me wrong. After all I was the author of the buggy/not performant previous version of this section of the codebase :P

core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java
```diff
@@ -119,53 +119,63 @@ public void start(Map<String, String> props) {
...
+                remaining = remaining(startPollEpochMilli, maxPollDuration);
+                continue;
+            }
+
+            if (records == null) {
```
valdar, line 142: I would rather initialize `records` with and empty list check at the end if it is empty to return `null` for readability, than  getting (little?) gain in performance...

valdar (changes requested):

orpiske:
> Hi @orpiske great work!

Thanks! And thanks for the review!

> 
> I was wondering which one of the contributions is predominant: better handling the remaining time or returning `null` if there are no results.

The biggest gain is related to avoiding the busy spin. On the previous code we were returning immediately from the [receiveNoWait](https://github.com/apache/camel-kafka-connector/blob/036cb9cd3053d71e4f1bb7e33abdd0bde013ee8d/core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java#L129) and so the loop was running thousands of time per poll. 

The change replaces the busy spin with a direct call to the receive, blocking for at most `maxPollDuration`.

> 
> Do you mind, since you already have the performance tests at hand, doing a quick comparison towards just returning `null` if the result set is empty, like: [valdar@2edcfc0](https://github.com/valdar/camel-kafka-connector/commit/2edcfc079134ec2742cac8e5fa625170f4141373)

I tried with returning the null as well. I have a flame graph with that test with [null available here](http://angusyoung.org/arquivos/tmp/profile-previous-with-null.svg).

It doesn't seem to affect the performance, but I changed it nonetheless because that way it complies with the [connect API](https://github.com/apache/kafka/blob/trunk/connect/api/src/main/java/org/apache/kafka/connect/source/SourceTask.java#L46-L60).

core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java
orpiske: > what is the point of this check? if the exchange is null, since you receive waiting for the whole remaining time it means you have finished the time so there is no point in updating it earlier and continue.

Great catch, man! I think you are right!

> 
> I would rather just check if exchange is not `null` and in that case do all the work except updating remaining time, i.e. something like:
> 
> ```java
> while(...) {
>     Exchange exchange = consumer.receive(remaining);
>     if(exchange != null) {
>     //whole code here
>     }
>     remaining = remaining(startPollEpochMilli, maxPollDuration);
> }
> ...
> ```

core/src/main/java/org/apache/camel/kafkaconnector/CamelSourceTask.java
orpiske: +1

valdar:
@orpiske thanks for the insight on this, I thought it was something similar but I just wanted to be sure.

valdar (approved):

oscerd (approved):

oscerd:
I believe after merging this we could cut a 0.5.0 release

orpiske:
> I believe after merging this we could cut a 0.5.0 release

+1 from me. I think we could start preparing that.

orpiske:
Looks like everything is alright: approvals OK, CI ok. So, merging this one.

PR #438, opened 2020-09-09
AWS SQS tests w/ SDK v2

orpiske:


oscerd (approved):

orpiske:
It's failing on deps only on JDK 11. This is very strange ... I will take a closer look tomorrow.

oscerd:
It was a glitch

Issue #279, opened 2020-06-17
Define the excluded Components list


oscerd:
After #278 has been merged, we need to set a list of excluded components

oscerd:
I'd like to revive the discussion. There are number of aws components that don't make sense as connector. For example kms,iam,eks,ecs etc. Also some others.
I'd like to limit the number of connectors to be able to speed up the release process and release often

@valdar @orpiske can you share your pov

orpiske:
> I'd like to revive the discussion. There are number of aws components that don't make sense as connector. For example kms,iam,eks,ecs etc. Also some others.
> I'd like to limit the number of connectors to be able to speed up the release process and release often
> 
> @valdar @orpiske can you share your pov

+1. I think this is a good idea and I agree with you that it can simplify release, maintenance and possibly even reduce any risk of extra work due to CVEs in dependencies. All in all, a welcome change, IMHO. 

It would be good - more as a safe proof for the future - to document our thought process in removing them. My thinking is that we might need to refer to that should the need arise. 

Lastly, we already have a lot of connectors. I don't know how you plan to do that, but it might be difficult for a single person to review them all. I volunteer to help with the effort to trim them.

oscerd:
Let me start by removing these:

camel-aws2-ec2
camel-aws2-ecs
camel-aws2-eks
camel-aws2-iam
camel-aws2-kms
camel-aws2-mq
camel-aws2-msk
camel-aws-ec2
camel-aws-ecs
camel-aws-eks
camel-aws-iam
camel-aws-kms
camel-aws-mq
camel-aws-msk

This components have only the producer side and are used to create, delete, list and describe stuff in AWS services- Lambda is similar, but you have the opportunity to invoke a function through the producer. So we should leave it. I'll open a first PR by removing these connectors.

oscerd:
I think we need to remove also 
- camel-mock
- camel-ref
- camel-robotframework (for testing)

I'll remove this and then I'll continue the investigation

omarsmak:
I think as well, camel-kafka doesn't make sense to be there since the whole thing is running based on kafka.

omarsmak:
As well, camel-debezium components I think in my opinion they don't need to be there as well since Debezium is originally a Kafka Connect connector.

oscerd:
Removing
- camel-kafka
- camel-debezium-*
- camel-dataset
- camel-dataset-test

oscerd:
Removing

- camel-digitalocean

[...]

PR #445, opened 2020-09-10
Cut down the connectors number

oscerd:
referes to #279


valdar:
I am not quite sure about removing `camel-kafka` ... isn't a legit usecase to read or send to _another_ Kafka cluster?

oscerd:
You won't use kafka connect for something like that iko

omarsmak:
For another kafka, that is a usecase for [MirrorMaker](https://kafka.apache.org/documentation/#basic_ops_mirror_maker) I think

oscerd:
Yes, if you need to do something like that, you'd not use kafka-connect. There are other tools.

PR #470, opened 2020-09-17
CAMEL-15496: Headers and properties support

rgannu:
The additional headers will be added on top of the existing headers
from the message.

The properties added are only basic AMQP properties as defined in the
{@link com.rabbitmq.client.AMQP.BasicProperties.Builder}. When the
message contains already these properties then will be considered and
these additional properties will be ignored.


oscerd (changes requested):
All this stuff is autogenerated, anything here will be overriden with a new build, so this stuff need to go in the component, considering the original review.

oscerd:
Anyway, if this will be merged upstream, it will be available on 3.6.0 and once we upgrade the connector will be aligned. This can be closed for now.

PR #475, opened 2020-09-22
Removed browse and class connector

oscerd:
Related to #279


oscerd:
@orpiske do you have time to investigate the elasticsearch pom problem while running the gh action? It seems to fail only on JDK11, maybe we could cache the repository and re-use it.

orpiske:
> @orpiske do you have time to investigate the elasticsearch pom problem while running the gh action? It seems to fail only on JDK11, maybe we could cache the repository and re-use it.

Yes I do. I will take a look at it.

PR #533, opened 2020-10-07
Clean up usages of org.junit.Assert

mureinik:
The project standard is to use `org.junit.jupiter.api.Assertions` for assertions, but there are several instances where the older `org.junit.Assert` is used, in what seems to be an oversight.

This patch cleans up those usages, and standardizes them to use `org.junit.jupiter.api.Assertions`, like the rest of the project.


orpiske (approved):

oscerd (approved):

oscerd:
Thanks

PR #536, opened 2020-10-07
Catalog: Rename the json file to reflect the id and connectors.properties content

### LINKED ISSUE #534: Catalog: Rename the json file to reflect the id and connectors.properties content
In connectors.properties we have the following:

camel-aws2-s3-source
camel-aws2-s3-sink

while in the connectors folder

camel-aws2-s3-kafka-sink-connector.json            
camel-aws2-s3-kafka-source-connector.json  

And inside these files the id is:

camel-aws2-s3-sink
camel-aws2-s3-source

so it probably makes sense to rename the json as something like <id>.json, to make the catalog easier to use.

Related to #430

oscerd:
Fixes #534


camel-kafka-connector-catalog/src/generated/resources/connectors/camel-ahc-sink.json
```diff
@@ -0,0 +1,162 @@
...
+		"camel.sink.endpoint.cookieHandler": {
+			"name": "camel.sink.endpoint.cookieHandler",
+			"description": "Configure a cookie handler to maintain a HTTP session",
+			"defaultValue": "null",
+			"priority": "MEDIUM"
+		},
+		"camel.sink.endpoint.headerFilterStrategy": {
+			"name": "camel.sink.endpoint.headerFilterStrategy",
+			"description": "To use a custom HeaderFilterStrategy to filter header to and from Camel message.",
+			"defaultValue": "null",
```
apupier, line 44: Isn't a problem to make the difference between the `null` as `String` and no default value?

I'm guessing here that it means there is no default value defined

oscerd: Yes, there is no default value when it's null. This is something coming from the how kafka connect define the options.

apupier: it means that we cannot define a default value which is the null String? or there are ways to escape it?

oscerd: I think we can improve the connector aurogeneration for this. @valdar cc

valdar: @oscerd it seems that we can use `ConfigDef.NO_DEFAULT_VALUE` kafka connect side to distinguish between no default value and `null` as default value. I am investigating if that information is available in the vanilla camel catalog that we process to generate connectors.

PR #546, opened 2020-10-08
source: better handling of headers of type Data

### LINKED ISSUE #544: source: better handling of headers of type Data
In the `CamelSourceTask`, if a Camel's header is of type date, it gets propagated as a string to Kafka with something like:

```java
if (value instanceof Date) {
    record.headers().addString(
        keyCamelHeader, 
        new SimpleDateFormat("yyyy-MM-dd").format(value));
}
```

this approach has some issues:

1. the consumer is not able to know the real type of the header as it is just a string
2. we lose precision as we remove the time of the day

As Kafka support date and time with `org.apache.kafka.connect.data.[Data|Time]`, we should probably avoid to use a string to represent dates however to fully translate a `java.util.Date` to the related Kafka types, we'd need to split it to multiple headers keys or using multiple headers for the same value.

I don't have enough knowledge at the moment about what could be the best solution but at least we should start by not losing the type.

lburgazzoli:
Fixes #544

### DIFF
--- a/core/src/main/java/org/apache/camel/kafkaconnector/CamelConnectorUtils.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/CamelConnectorUtils.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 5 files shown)


lburgazzoli:
This is a partial impl that covers the `Date` time being preserved

oscerd (approved):

lburgazzoli:
timestamp implemented @oscerd m,ind having a look ?

oscerd (approved):

PR #548, opened 2020-10-09
Catalog: expose an API for consuming connectors catalog metadata and options

### LINKED ISSUE #542: Catalog: expose an API for consuming connectors catalog metadata and options
I imagine having in camel-language-server pom, a dependency on the catalog, something like:
```
<dependency>
  <groupId>org.apache.camel.kafkaconnector</groupId>
  <artifactId>camel-kafka-connector-catalog</artifactId>
</dependency>
```

and then in Camel Language Server code, having an API Method which is returning the loaded Model as a Java Class, something like:

```
CamelKafkaConnectorCatalog catalog = new CamelKafkaConnectorCatalog();
List<CamelKafkaConnectorModel> connectors = catalog.getConnectors();

```

_Originally posted by @apupier in https://github.com/apache/camel-kafka-connector/issue_comments/705513426_

Related to #430

oscerd:
Fixes #542

### DIFF
--- a/camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
+++ b/camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
@@ -17,6 +17,7 @@
 package org.apache.camel.kafkaconnector.catalog;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -42,6 +43,9 @@ public class CamelKafkaConnectorCatalog {
     static List<String> connectorsName = new ArrayList<String>();
     static Map<String, CamelKafkaConnectorModel> connectorsModel = new HashMap<String, CamelKafkaConnectorModel>();
     private static final Logger LOG = LoggerFactory.getLogger(CamelKafkaConnectorCatalog.class);
+    private static final String CONNECTORS_DIR = "src/generated/resources/connectors";
+    private static final String DESCRIPTORS_DIR = "src/generated/resources/descriptors";
+    private static final String CONNECTORS_PROPERTIES = "connectors.properties";
 
     public CamelKafkaConnectorCatalog() {
         initCatalog();
@@ -55,7 +59,7 @@ public class CamelKafkaConnectorCatalog {
     }
 
     private void initCatalog() {
-        try (FileInputStream input = new FileInputStream("src/generated/resources/descriptors/connectors.properties")) {
+        try (FileInputStream input = new FileInputStream(DESCRIPTORS_DIR + File.separator + CONNECTORS_PROPERTIES)) {
 
             BufferedReader reader = new BufferedReader(new InputStreamReader(input));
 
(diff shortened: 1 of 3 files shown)


camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
```diff
@@ -0,0 +1,138 @@
...
+            model.add(singleModel);
+        }
+        return model;
+    }
+
+    public List<String> getConnectorsName() {
+        return connectorsName;
+    }
+
+    public static Map<String, CamelKafkaConnectorModel> getConnectorsModel() {
```
apupier, line 135: why this method is static?

If accessed before an instanciation of the class happened, it will return an empty value.

camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
```diff
@@ -0,0 +1,138 @@
...
+            LOG.error("Cannot find file: {}", e.getMessage(), e);
+        } catch (IOException e) {
+            LOG.error("IO Exception: {}", e.getMessage(), e);
+        }
+    }
+
+    public String getConnectorAsJson(String connectorName) {
+        String result = null;
+        try {
+            result = Files.lines(Paths.get(CONNECTORS_DIR + File.separator + connectorName + ".json")).parallel() // for
```
apupier, line 90: I think that accessing this way doesn't work with a bundled jar. It is working only in development mode with an exploded folder. Next to use some getResourceAsStream methods

I will have a look to provide PR

camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
oscerd: It's a mistake.

camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
oscerd: We can also use the catalog helper.

oscerd: From core catalog.

camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
apupier: ok, I will provide a PR to remove the static

oscerd: It's already done and merged

PR #555, opened 2020-10-09
Remove static access to CamelKafkaConnectorCatalog.getConnectorsModel()

apupier:
the class needs to be instantiated to have model initialized


apupier:
already done

PR #574, opened 2020-10-12
573 allow to add remove elements from catalog dynamically through api

apupier:


camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
```diff
@@ -83,7 +83,7 @@ private void initCatalog() {
         }
     }
 
-    public String getConnectorAsJson(String connectorName) {
+    private String loadConnectorAsJsonFromEmbeddedCatalog(String connectorName) {
```
oscerd, line 86: I wouldn't change this name. It doesn't add so much.

camel-kafka-connector-catalog/src/test/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalogTest.java
```diff
@@ -20,30 +20,30 @@
 import java.util.Map;
 
 import org.apache.camel.kafkaconnector.model.CamelKafkaConnectorModel;
+import org.apache.camel.kafkaconnector.model.CamelKafkaConnectorOptionModel;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
oscerd, line 27: Please don't import with * operator

oscerd (commented):
Just two things: the method name is not very important. The import could be easier and just use the needed classes.

camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
apupier: The added difference is that the method is loading from embedded catalog, it is not able to provide the json from the cached connector that are dynamically added. With the name "getConnectorAsJson", I'm expecting to have also the dynamically added connector to be provided.

camel-kafka-connector-catalog/src/test/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalogTest.java
apupier: I was only using the defaults of Eclipse projects but fine to follow another rule.

camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
oscerd: Sounds good them

oscerd:
LGTM. Let me merge.

PR #579, opened 2020-10-13
Jenkinsfile: Changed the JDK naming to reflect new INFRA approach

oscerd:


oscerd:
It seems the update is not in place yet

Issue #581, opened 2020-10-13
Strimzi 0.20: Lets test early with Release candidate


oscerd:
The first one, is already out. We can do this on camel-master
https://github.com/strimzi/strimzi-kafka-operator/releases/tag/0.20.0-rc1

orpiske:
@oscerd if you want, I can prepare this.

oscerd:
Sure, but I'd like to wait for the final release, so maybe we could have a profile for testing with release candidates? Not sure. This would probably require updating to kafka 2.6.0 too

orpiske:
I'm trying the following:

orpiske:
I am not sure if the `-Dkafka.version=2.6.0` yet will actually manage to override the version, but the rest should definitely run the tests against Strimzi 0.20 rc1. I am running a build here and will let you know soon.

oscerd:
Thanks, sounds good. We can do this and if it works we can create a profile to be enabled with a gh action for example.

orpiske:
That sounds like a good plan +1.

orpiske:
@oscerd looking at the GH actions, it does not seem to have run. I believe I need to cherry-pick the action into master.

oscerd:
yes

orpiske:
Worked after cherry-picking. Therefore, marking this one as closed.

PR #583, opened 2020-10-13
Added a new profile so that we can test w/ newer Strimzi versions

orpiske:
This should do the trick for #581:

```
mvn -Pstrimzinext -DskipIntegrationTests=false clean verify
```

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -506,6 +506,18 @@
 
     <profiles>
         <profile>
+            <id>strimzinext</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <properties>
+                <kafka.version>2.6.0</kafka.version>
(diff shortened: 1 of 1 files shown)


oscerd (approved):

oscerd:
Ok. We now need to create a gh action for this :-)

orpiske:
> Ok. We now need to create a gh action for this :-)

I think I can squeeze that into this PR. I'm taking a look at that :)

orpiske:
So, I added a new action based on Sync Camel Kafka Connector camel-master Branch. 

It's very similar but contains the following differences: 
- Builds regularly at 3 AM instead of at 2 AM
- Do not send PRs after the build

I am not proficient w/ the GH actions, so just let me know if there's something I can do better.

oscerd (approved):
LGTM

PR #586, opened 2020-10-14
Changes to documents - change version from 0.3.0-SNAPSHOT, give the right location of the properties file within docs/examples

cunningt:
I've been going through the examples and noticed a few things in the local doc 
- was using 0.3.0-SNAPSHOT
- was referring to connector properties files in config or examples - it looks like they exist now in docs/examples
- added formatting to plugin.path


oscerd (approved):

oscerd:
Thanks a lot

PR #587, opened 2020-10-14
Added GH actions to run the camel-master build using the Strimzi Next profile

orpiske:


orpiske:
No code change and already merged, so safe to merge.

PR #588, opened 2020-10-14
Avoid usage of static for Camel Kafka Connector catalog

apupier:
it will allow to use several version of the catalog in the same JVM.


oscerd (approved):
Not a problem, but why do you need multiple catalogs in the same JVM? The embedded catalog is static by definition.

apupier:
Given that I added the possibility to add/remove connector, the Catalog is dynamic and is not handling only embedded connector definitions.

for right now, to be able to write tests with different catalog more easily

for the future:
- to allow handle several version of Camel Kafka Connector in the tooling
- to let users provide their own connector definition

PR #608, opened 2020-10-17
Apply the Strimzi next profile only to the connector build

orpiske:


orpiske:
Only applies to the GH action, fixing trying to build Camel w/ the wrong profile.

PR #638, opened 2020-10-23
Decouple the Elasticsearch client from the service to simplify moving the services to Camel core

orpiske:


oscerd (approved):
I think @valdar is cutting the release. So lets wait for merging this.

orpiske:
@oscerd sure thing, thanks for the heads up!

Issue #697, opened 2020-11-10
Camel Kafka Connector Catalog throws NPE on Windows


apupier:
see https://github.com/camel-tooling/camel-lsp-client-vscode/issues/472

```
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.boot.loader.MainMethodRunner.run(MainMethodRunner.java:49)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:107)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:58)
	at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:88)
Caused by: java.lang.NullPointerException
	at java.base/java.io.Reader.<init>(Reader.java:167)
	at java.base/java.io.InputStreamReader.<init>(InputStreamReader.java:72)
	at org.apache.camel.kafkaconnector.catalog.CamelKafkaConnectorCatalog.initCatalog(CamelKafkaConnectorCatalog.java:65)
	at org.apache.camel.kafkaconnector.catalog.CamelKafkaConnectorCatalog.<init>(CamelKafkaConnectorCatalog.java:52)
	at com.github.cameltooling.lsp.internal.catalog.util.CamelKafkaConnectorCatalogManager.<init>(CamelKafkaConnectorCatalogManager.java:23)
	at com.github.cameltooling.lsp.internal.CamelTextDocumentService.<init>(CamelTextDocumentService.java:94)
	at com.github.cameltooling.lsp.internal.CamelLanguageServer.<init>(CamelLanguageServer.java:53)
	at com.github.cameltooling.lsp.internal.Runner.main(Runner.java:65)
	... 8 more
[Info  - 11:24:59 am] Connection to server got closed. Server will restart.
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #698, opened 2020-11-10
Avoid NPE on Windows for Catalog initialization #697

apupier:
path inside bundles are always using "/", it is a resource path, not a
file path. i tis then not using a file separator.

### DIFF
--- a/camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
+++ b/camel-kafka-connector-catalog/src/main/java/org/apache/camel/kafkaconnector/catalog/CamelKafkaConnectorCatalog.java
@@ -17,7 +17,6 @@
 package org.apache.camel.kafkaconnector.catalog;
 
 import java.io.BufferedReader;
-import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


oscerd:
This was introduced here: https://github.com/apache/camel-kafka-connector/commit/0c6871f86a9d71d47bd7144c2e636509d0a80fac

oscerd (approved):

apupier:
> This was introduced here: [0c6871f](https://github.com/apache/camel-kafka-connector/commit/0c6871f86a9d71d47bd7144c2e636509d0a80fac)

not exactly as it was not working at all with a built jar before this commit. It has never worked on Windows with a built jar.

oscerd:
In the beginning this was just a PoC to provide something, it wasn't developed to be a jar, but just to return something to consume. Also Windows is not really our main focus.

apupier:
> In the beginning this was just a PoC to provide something, it wasn't developed to be a jar,

it is consumed as a jar. It must work as a jar.
Having the Proof of concept, even if not working as a jar, was a great start. Now, we can incrementally improve it to support more cases.

> Also Windows is not really our main focus.

for the catalog, it is an important target given that it is used during development.
I do not have figures of OS usage for VS Code side but for Eclipse Fuse Tooling, where it is also used, 80% is Windows.

PR #723, opened 2020-11-19
File language and core languages in general are not working 

### LINKED ISSUE #720: File language and core languages in general are not working
Also for 0.6.0

camel.sink.endpoint.keyName=${date:now:yyyyMMdd-HHmmssSSS}-${exchangeId}

It won't be substituted anyway. We can workaround the problem by using headers where possible.

@lburgazzoli I think this is coming from the introduction of templating

oscerd:
Fixes #720 

This will go in 0.7.0


oscerd:
Fix is good. We need new camel snapshot

PR #752, opened 2020-11-29
Set JMS clients scope runtime to include required dependencies in the package

fvaleri:
Not sure if there is a better way to do this, but it works and it is much more convenient. What do you think?

### DIFF
--- a/connectors/camel-kafka-connector-fix-dependencies.properties
+++ b/connectors/camel-kafka-connector-fix-dependencies.properties
@@ -21,10 +21,7 @@
 # Global dependencies included in all modules (except the ones in the list)
 global=org.apache.camel.kafkaconnector:camel-kafka-connector
 
-# we add a default connection factory maven dependency (variables can be used as ${varname})
-camel-sjms2=org.apache.activemq:activemq-client,org.apache.activemq:artemis-jms-client
+# we add a default connection factory maven variables can be used as ${varname}
+camel-sjms2=org.apache.activemq:activemq-client::runtime,org.apache.activemq:artemis-jms-client::runtime
 exclude_camel-sjms2=
 additional_properties_camel-sjms2=camel.component.sjms2.connection-factory=#class:org.apache.activemq.ActiveMQConnectionFactory,camel.component.sjms2.connection-factory.brokerURL=tcp://localhost:61616
-
-# issue 259: https://github.com/apache/camel-kafka-connector/issues/259
-camel-elasticsearch-rest=org.slf4j:log4j-over-slf4j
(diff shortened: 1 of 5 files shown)


oscerd (changes requested):
This will make all the connectors packaging jms client and artemis jms client, even where not needed at all

fvaleri:
Yeah, this was my concern. Is there a better way to do this automatically?

oscerd:
The only way I see is making the deps explicit in the connector pom of the single connectors.

fvaleri:
This commit should be better.

oscerd (approved):
Seems ok. Lets see the other reviewers @valdar and @orpiske

valdar:
Mmm should have been enough to have them as dependencies... not sure why the runtime scope is needed, pretty sure they ended up in the packaged zip at sone point...

oscerd:
Both of them are declared in parent pom as test dependencies https://github.com/apache/camel-kafka-connector/blob/master/parent/pom.xml#L292-L302

So when it's time to build the packaged connector they will be excluded, because of the scope.

fvaleri:
> Both of them are declared in parent pom as test dependencies https://github.com/apache/camel-kafka-connector/blob/master/parent/pom.xml#L292-L302
> 
> So when it's time to build the packaged connector they will be excluded, because of the scope.

Yeah. They were in the final package in 0.3, but not anymore. Adding the scope to the fix-dependencies property file seems to work fine and the same technique could be used for other connectors as needed.

orpiske:
@oscerd @valdar I believe you both have better understanding of the ramifications of this than I do. At first sight I have no objections with the approach, but I would kindly ask for one more thing. 

Do we have to update the [missing dependencies part](https://github.com/apache/camel-kafka-connector/blob/master/docs/modules/ROOT/pages/troubleshooting.adoc) of the troubleshooting guide as part of this? 

My understanding is that we do need to update the note there. Therefore, I would probably just put a note saying "From version 0.x.x and forward the JMS dependencies are packaged and this should not be a problem. Should you need to use other JMS clients, you need to repackage or use the legacy/regular/etc JMS component".

fvaleri:
> @oscerd @valdar I believe you both have better understanding of the ramifications of this than I do. At first sight I have no objections with the approach, but I would kindly ask for one more thing.
> 
> Do we have to update the [missing dependencies part](https://github.com/apache/camel-kafka-connector/blob/master/docs/modules/ROOT/pages/troubleshooting.adoc) of the troubleshooting guide as part of this?
> 
> My understanding is that we do need to update the note there. Therefore, I would probably just put a note saying "From version 0.x.x and forward the JMS dependencies are packaged along w/ sjms2 and this should not be a problem. Should you need to use other JMS clients, you need to repackage or use the legacy/regular/etc JMS component".

+1 for this, thanks @orpiske.

In general I think that trying to avoid the need to repackage is the best strategy for a great user experience.

Let me know if you want me to add this note to the commit.

orpiske:
@fvaleri thanks. Yes, please. I think it would be better to have that note added now that the discussion is fresh. Thanks!

valdar:
tnx @fvaleri !

PR #789, opened 2020-12-15
Camel-Master updated to 3.7.0 staging

oscerd:
@orpiske @valdar with this we should be done for releasing 3.7.0


oscerd:
Once Camel 3.7.0 will be released we just need to rebase master on camel-master and we should be ok.

orpiske:
@oscerd that's great news. Looking forward to that one!

PR #805, opened 2020-12-17
Support service account authentication

jmerljak:
Service account authentication support for google-calendar-stream connector.
Depends on https://github.com/apache/camel/pull/4783


oscerd (commented):
This stuff is autogenerated. You'll need to wait for release 3.8.0 of camel. Once it will be released in the future, we'll update. This cannot work with 3.7.0 and with the current master. Thanks anyway.

PR #828, opened 2021-01-07
Update try-it-out-on-kubernetes procedure

fvaleri:
Fixed an error with Strimzi release variable. Small improvements to the command line procedure.


oscerd (changes requested):
Trying ckc with strimzi i don't think it requires 3 nodes. A single node deployment is enough i guess

fvaleri:
Yes, but I'm setting 3 nodes because it's a better setup in case you want to scale up and start doing serious business. Otherwise you would need to recreate KafkaConnect internal topics to have a more reliable setup with 3 replicas.

oscerd (approved):

oscerd:
Fine. But while trying you won't do serious business

Issue #159, opened 2020-04-28
Camel components secret options are showed in logs


oscerd:
If you set an option like for example an AWS secret key in one of the connector configuration it will be showed up in the log in clear.

valdar:
Do we have some facility at Camel level to deal with this? I suspect that we must search for such an option in Kafka connect framework.

davsclaus:
Yeah see how we do something like that in camel-main or the santize uri method

org.apache.camel.util.URISupport#sanitizeUri
org.apache.camel.main.BaseMainSupport#SENSITIVE_KEYS

oscerd:
Should be ok now.

oscerd:
Let's keep it open until we update master to 3.8.0

oscerd:
Done

PR #830, opened 2021-01-07
(backport 0.7.x) Fix issue #159: camel components secret options are showed in logs

orpiske:


oscerd (approved):

oscerd (commented):
I think you need to change camel-dependencies version too, here and on camel-master

orpiske:
> I think you need to change camel-dependencies version too, here and on camel-master

Oh, good point! I also think we need to adjust this: https://github.com/apache/camel-kafka-connector/blob/camel-kafka-connector-0.7.x/pom.xml#L31 

On 0.7.x, it is versioned as 0.8.

orpiske:
BTW, I am going to separate the upgrade change from this/these, otherwise it's going to be messy to figure out what went where. We can leave these ones open and I'll rebase them after the fact.

orpiske:
Ok, this one is done and we should be good to go. Merging and starting the work on fixing the other PR ...

PR #832, opened 2021-01-07
(backport) Bump camel dependency to use camel 3.7.1-SNAPSHOT

orpiske:


orpiske:
Camel dependencies is already set to 3.7.1-SNAPSHOT. This just set the camel.version variable to the same.

oscerd (approved):

Issue #835, opened 2021-01-08
Error reporting configuration incorrectly handled


orpiske:
The sink connector does not respect the [new error reporting](https://kafka.apache.org/26/documentation.html#connect_errantrecordreporter) configuration from Kafka Connect. As of Kafka 2.6, it seems, we should be using the Errant Record Reporter to ensure that failed records are put into the DLQ. 

One way to verify this issue is to use and invalid JMS broker address in the [CamelSinkJMSITCase](https://github.com/apache/camel-kafka-connector/blob/camel-kafka-connector-0.7.0/tests/itests-sjms2/src/test/java/org/apache/camel/kafkaconnector/sjms2/sink/CamelSinkJMSITCase.java#L71) test. This causes a failure in the component which, eventually, leads to a [ConnectionException](https://github.com/apache/camel-kafka-connector/blob/camel-kafka-connector-0.7.0/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java#L178) and lost records. 

Here's a snippet taken from one of my logs as an example:

```
Caused by: org.apache.kafka.connect.errors.ConnectException: Exchange delivery has failed!
        at org.apache.camel.kafkaconnector.CamelSinkTask.put(CamelSinkTask.java:185) ~[camel-kafka-connector-0.8.0-SNAPSHOT.jar:0.8.0-SNAPSHOT]
        at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:560) ~[connect-runtime-2.6.0.jar:?]
        ... 10 more
Caused by: org.apache.camel.ResolveEndpointFailedException: Failed to resolve endpoint: sjms2://ckc.queue due to: javax.jms.JMSException: No Provider scheme specified.
```

The message is not sent to the DLQ in this case, which violates the expected behavior.

orpiske:
Fixed. Should be available on 0.7.1 and 0.8.0.

oscerd:
@orpiske maybe we could add a little example of this scenario in ckc-examples repository, so we could point user there

orpiske:
> @orpiske maybe we could add a little example of this scenario in ckc-examples repository, so we could point user there

+1. That's a good idea. I will add an example covering this item.

PR #836, opened 2021-01-08
Ensure the DLQ configuration from Kafka Connect is correctly handled (issue #835)

orpiske:


orpiske:
We'll also need this on 0.7.x.

oscerd (approved):

Issue #844, opened 2021-01-11
Camel-Master: SJMS2 it test is failing


oscerd:
@orpiske can you have a look? There were some updates on the component on 3.8.0-SNAPSHOT.

Thanks.

orpiske:
Sure, I'll take a look!

oscerd:
seems to be done.

PR #845, opened 2021-01-11
Avoid slashes in sjms2 URLs

orpiske:
This should do the trick for issue #844


oscerd (approved):

orpiske:
The failure is on another component, so I am merging this one.

Issue #816, opened 2020-12-24
Support camel dataformat configuration for marshaller/unmarshaller


luigidemasi:
there are some dataformats that support configuration parameters, but at the moment is just possible to specify the dataformat id in `camel.source.marshal` and `camel.source.unmarshal`

valdar:
So based on https://github.com/apache/camel-kafka-connector/pull/855 this should have already worked, but a bug in camel was discovered, reported and fixed see https://issues.apache.org/jira/browse/CAMEL-16032.

This can be closed once we use a version of camel that has the fix.
NOTE: this test must be updated https://github.com/apache/camel-kafka-connector/commit/712917ea9274feaf13bb5f4eeb2971b4f0c6e234

PR #855, opened 2021-01-12
Support camel dataformat configuration for marshaller/unmarshaller #816

luigidemasi:
this pr allow the configuration of a dataformat directly in properties file, something like:

```
camel.source.unmarshal=fhirJson
camel.source.unmarshal.fhirJson.fhirVersion=R4
camel.source.unmarshal.fhirJson.prettyPrint=true
```

### DIFF
--- a/core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectDataformat.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectDataformat.java
@@ -16,6 +16,51 @@
  */
 package org.apache.camel.kafkaconnector.utils;
 
+import java.util.Map;
+import java.util.Properties;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.apache.camel.CamelContext;
+import org.apache.camel.model.DataFormatDefinition;
+import org.apache.camel.model.RouteDefinition;
+import org.apache.camel.model.dataformat.ASN1DataFormat;
+import org.apache.camel.model.dataformat.Any23DataFormat;
+import org.apache.camel.model.dataformat.AvroDataFormat;
+import org.apache.camel.model.dataformat.Base64DataFormat;
+import org.apache.camel.model.dataformat.BeanioDataFormat;
+import org.apache.camel.model.dataformat.BindyDataFormat;
+import org.apache.camel.model.dataformat.CBORDataFormat;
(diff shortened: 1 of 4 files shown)


oscerd (approved):
Looks good to me. Thanks a lot. @valdar cc

oscerd:
Can you fix the codestyle?

core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectDataformat.java
```diff
@@ -35,6 +80,147 @@ public CamelKafkaConnectDataformatKind getDataformatKind() {
...
+
+    public static DataFormatDefinition configureDataformat(CamelContext context, String dataFormatName, Predicate<String> keyFilter, Properties camelProperties) {
+        Map<String, Object> dataFormatProps = camelProperties.stringPropertyNames()
+                .stream()
+                .filter(keyFilter)
+                .collect(Collectors.toMap(key -> key.substring(key.lastIndexOf(".") + 1), key -> camelProperties.get(key)));
+
+        DataFormatDefinition dataformat;
+
+        switch (dataFormatName) {
```
valdar, line 115: I hope this switch case could be avoided by something on the line of `context.createDataFormat(dataFormatName)`

valdar (changes requested):
I was under the general impression that this usecase should have worked already...
Give me some more time to investigate further.

luigidemasi:
> I was under the general impression that this usecase should have worked already using something like:
> 
> ```
> camel.dataformat.fhirJson.fhirVersion=R4
> camel.dataformat.fhirJson.prettyPrint=true
> ```
> 
> see https://github.com/apache/camel-kafka-connector/blob/master/core/src/test/java/org/apache/camel/kafkaconnector/DataFormatTest.java#L128
> Give me some more time to investigate further.

It should work but it doesn't, if you look at the test, you disable validation and then assert that the validation is enabled.
I think it's a bug on camel-main atuoconfiguration that doesn't bind the dataformat in the registry and a new one is created when the route starts. Opened an issue for this: [CAMEL-16032](https://issues.apache.org/jira/browse/CAMEL-16032)

valdar:
Thanks a lot for the effort in seeking the root cause of this and fixing it @luigidemasi ! 
I close this and the effort is tracked here: https://github.com/apache/camel-kafka-connector/issues/816

PR #860, opened 2021-01-13
Removed unused AWS common modules that was moved to Camel Core

orpiske:


oscerd (approved):

orpiske:
The change is ok, but failing due to DH pull limits. Merging and checking later on Apache CI.

Issue #879, opened 2021-01-18
HDFS tests running suspiciously flaky on Github actions


orpiske:
More than once the HDFS tests have caused the github actions to hang and run for a very long time. This could be caused by docker pull limits, but also could be caused by some resource usage limits on Github itself - this test is heavy.

orpiske:
Likely caused by a conflict with Kafka 2.7.0:

```
021-01-18 18:06:58,937 [tp1105063017-88] WARN org.eclipse.jetty.server.HttpChannel - null
java.lang.NoSuchMethodError: 'org.eclipse.jetty.util.Attributes org.eclipse.jetty.util.Attributes.unwrap(org.eclipse.jetty.util.Attributes)'
	at org.eclipse.jetty.server.Request.recycle(Request.java:1891) ~[jetty-server-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.server.HttpChannel.recycle(HttpChannel.java:309) [jetty-server-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.server.HttpChannelOverHttp.recycle(HttpChannelOverHttp.java:81) [jetty-server-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.server.HttpConnection.onCompleted(HttpConnection.java:436) ~[jetty-server-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.server.HttpChannel.onCompleted(HttpChannel.java:743) ~[jetty-server-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:363) [jetty-server-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:273) [jetty-server-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311) [jetty-io-9.4.33.v20201020.jar:9.4.33.v20201020]
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105) [jetty-io-9.4.33.v20201020.jar:9.4.33.v20201020]
[...]

PR #893, opened 2021-01-19
Fix HDFS tests incompatibilities

orpiske:
FYI: it's likely we'll have to do the same for the connector deps, but I'll need more time to verify it. I'll send a separate PR for that, if it's the case.


orpiske:
This should fix #879

oscerd (approved):

PR #918, opened 2021-01-25
Java 14 changes

orpiske:
I am opening this draft PR with some of the changes I had to do in order to build the project with Java 14. 

Let's not merge this one yet ... I think it needs some review because of several plugin upgrades.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -56,8 +56,8 @@
 
 
         <mycila-license-version>3.0</mycila-license-version>
-        <gmavenplus-plugin-version>1.6.2</gmavenplus-plugin-version>
-        <groovy-version>2.5.8</groovy-version>
+        <gmavenplus-plugin-version>1.9.0</gmavenplus-plugin-version>
+        <groovy-version>2.5.10</groovy-version>
 
         <itest.zookeeper.container.image>quay.io/strimzi/kafka:0.21.1-kafka-2.7.0</itest.zookeeper.container.image>
         <itest.strimzi.container.image>quay.io/strimzi/kafka:0.21.1-kafka-2.7.0</itest.strimzi.container.image>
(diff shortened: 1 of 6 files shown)


parent/pom.xml
```diff
@@ -56,8 +56,8 @@
 
 
         <mycila-license-version>3.0</mycila-license-version>
-        <gmavenplus-plugin-version>1.6.2</gmavenplus-plugin-version>
-        <groovy-version>2.5.8</groovy-version>
+        <gmavenplus-plugin-version>1.9.0</gmavenplus-plugin-version>
+        <groovy-version>2.5.10</groovy-version>
```
lburgazzoli, line 60: For my understanding, why not using groovy 3.x ?

oscerd (approved):
I think that we can try to upgrade groovy to 3.x

orpiske:
@lburgazzoli @oscerd I tried to be conservative with the upgrades, but if you are comfortable with that, we can try going with Groovy 3.x. I will modify this later today and run some builds for testing.

lburgazzoli:
> @lburgazzoli @oscerd I tried to be conservative with the upgrades, but if you are comfortable with that, we can try going with Groovy 3.x. I will modify this later today and run some builds for testing.

I think Camel is already on Groovy 3.x so it should be ok, I'm also using Groovy 3.x since long time in camel-k-runtime and so fare everything seems good

apupier:
what about adding the JDK 14 in th ebuild matrix? it will allow to see on this PR directly if it is building fine and also detect regressions.

orpiske:
> what about adding the JDK 14 in th ebuild matrix? it will allow to see on this PR directly if it is building fine and also detect regressions.

+1 from me. I'll do that.

oscerd (approved):

oscerd:
Seems to be ok.

orpiske:
> Seems to be ok.

I agree. It looks OK so far. I have run a few additional tests on a CI I have @ home and everything seemed to run OK. I'll remove the draft flag on the PR.

claudio4j:
After Java 9 there is the `release` parameter, shouldn't use it for Java 14 in the `maven-compiler-plugin` configuration section, instead of `source/target` ? Example:
```
<configuration>
    <release>14</release>
</configuration>
```

oscerd:
No, because we need to support java 8 as base. Pointing directly java 14 will drop jdk8

oscerd:
@orpiske if you are ok, we can merge.

orpiske:
> @orpiske if you are ok, we can merge.

Thanks @oscerd, @lburgazzoli, @apupier and @claudio4j for the review and comments. I'll merge this and we can fine tune later if needed.

PR #925, opened 2021-01-27
Fix table doc and rename connecotr word

claudio4j:


oscerd (changes requested):
I guess the header file is used somewhere in the code. Can you double check? Thank you!

claudio4j:
The wrong file name was checked with `grep -Hr connecotr .` and is not used anywhere. The correct file name is used in AbstractCamelKafkaConnectorMojo.java.

oscerd:
Thanks

PR #926, opened 2021-01-28
Removes the daily JDK 14 github action because it is already being run on the default CI action

orpiske:


orpiske:
@apupier here it is. Thanks for the heads up!

Issue #873, opened 2021-01-15
add itests for camel-cxf-kafka-connector


ffang:

ffang:
Hi @valdar, @oscerd, @orpiske ,

Happy Friday!

I just sent a PR for the cxf-itests, including both Source and Sink CXF connector itests case.

And I have a few questions, I will add comments in the PR near the code piece I have questions. Please take a look at this PR!

Thanks!
Freeman

orpiske:
Thanks @ffang! I will take a look at it :)

valdar:
I think some itest for cxf are there (I fixed some myself for camel 3.9.0 release) so I would close this.

PR #940, opened 2021-01-29
#873 initial cxf Source/Sink connectors test

ffang:


valdar (changes requested):
I marked this as request changes to keep track of the suggestions @orpiske gave.

ffang:
@orpiske revised accordingly per your feedback. Thanks!

tests/itests-cxf/src/test/java/org/apache/camel/kafkaconnector/cxf/sink/GreeterImpl.java
```diff
@@ -15,16 +15,16 @@
...
 
-/**
- * A producer of test messages
- */
-public interface TestMessageProducer<T> {
-    Map<String, String> messageHeaders(T text, int current);
-    T testMessageContent(int current);
-    void produceMessages() throws ExecutionException, InterruptedException;
+
+public class GreeterImpl extends org.apache.hello_world_soap_http.GreeterImpl {
```
aimiebell, line 23: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names shouldn’t shadow simple name of superclass (NM_SAME_SIMPLE_NAME_AS_SUPERCLASS)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-shouldn-t-shadow-simple-name-of-superclass-nm-same-simple-name-as-superclass).
This class has a simple name that is identical to that of its superclass, except that its superclass is in a different package (e.g., alpha.Foo extends beta.Foo). This can be exceptionally confusing, create lots of situations in which you have to look at import statements to resolve references and creates many opportunities to accidentally define methods that do not override methods in their superclasses.

oscerd: Thanks, but this is how cxf services are usually generated.

orpiske (approved):
Thanks @ffang! LGTM.

ffang:
Merged, thanks guys!

PR #970, opened 2021-02-03
Cleanup sink test cases

orpiske:


oscerd (approved):

oscerd:
Can you create the same PR for camel-master? The backport action will work only if the branch is on the same repository, it won't work from fork. I cannot find workaround for this, it is a security enforcement from ASF.

orpiske:
> Can you create the same PR for camel-master? The backport action will work only if the branch is on the same repository, it won't work from fork. I cannot find workaround for this, it is a security enforcement from ASF.

Yes I can. I'll create that once this one is in good shape :)

PR #1006, opened 2021-02-12
Cleanup the check state logic on the KafkaConnectEmbeddedService

orpiske:


oscerd (approved):

oscerd:
I think this weekend we'll start the cut for 3.7.2.

We'll need to align camel-master to all the commits you've done @orpiske and then merge on master maybe on Monday, so we could work on the missing bit for 0.8 and release next weekend.

orpiske:
Ack @oscerd. Soon I'll open a PR against camel-master with all the changes I did during the last days. The code may need some massaging, so it would be better to start soon and then get to a clean state in time.

PR #1017, opened 2021-02-16
(backport) Delay starting the producer on the syslog test to avoid failing due to port not being available

orpiske:


oscerd (approved):

orpiske:
First test went ok, but I want to run one more time to make sure it's fully resolved.

PR #1021, opened 2021-02-17
Make sure CXF tests are included in the test run

orpiske:


oscerd (approved):

orpiske:
The failure is not caused by the change, so I am merging this one and sending the cleanup on a separate pr.

PR #1052, opened 2021-02-23
Convert Struct to Map

### LINKED ISSUE #1046: Add a SMT/Converter to convert JSON to Map in the Core
There are components/connectors that need the message body to be in form of Map, for example Camel Stitch, Camel SQL .. etc. Hence if you have a JSON data message from Kafka, e.g:
```
{
    "name" : "joe doe",
    "id": 12
}
``` 
The SMT should be able to convert this to Map datatype. e.g:
```
map.get("name");
map.get("id");
```

omarsmak:
Fixes #1046 

With PR, the Sink will automatically convert any KafkaConnect Struct data type to Map. Theoretically Camel components doesn't need any data type of struct since it is Kafka Connect specific data types and hence just convert it to map that is propagated to Camel components.
It converts body, key and headers

### DIFF
--- a/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
@@ -18,6 +18,7 @@ package org.apache.camel.kafkaconnector;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 
@@ -33,7 +34,9 @@ import org.apache.camel.support.DefaultExchange;
 import org.apache.camel.util.StringHelper;
 import org.apache.kafka.connect.data.Decimal;
 import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.header.Header;
 import org.apache.kafka.connect.sink.ErrantRecordReporter;
 import org.apache.kafka.connect.sink.SinkRecord;
@@ -110,7 +113,7 @@ public class CamelSinkTask extends SinkTask {
             final String headersRemovePattern = config.getString(CamelSinkConnectorConfig.CAMEL_CONNECTOR_REMOVE_HEADERS_PATTERN_CONF);
             mapProperties = config.getBoolean(CamelSinkConnectorConfig.CAMEL_CONNECTOR_MAP_PROPERTIES_CONF);
             mapHeaders = config.getBoolean(CamelSinkConnectorConfig.CAMEL_CONNECTOR_MAP_HEADERS_CONF);
-            
+
             CamelContext camelContext = new DefaultCamelContext();
             if (remoteUrl == null) {
                 remoteUrl = TaskHelper.buildUrl(camelContext,
@@ -175,8 +178,8 @@ public class CamelSinkTask extends SinkTask {
             TaskHelper.logRecordContent(LOG, loggingLevel, record);
 
             Exchange exchange = new DefaultExchange(producer.getCamelContext());
-            exchange.getMessage().setBody(record.value());
-            exchange.getMessage().setHeader(KAFKA_RECORD_KEY_HEADER, record.key());
+            exchange.getMessage().setBody(convertValueFromStruct(record.valueSchema(), record.value()));
+            exchange.getMessage().setHeader(KAFKA_RECORD_KEY_HEADER, convertValueFromStruct(record.keySchema(), record.key()));
 
             for (Header header : record.headers()) {
                 if (header.key().startsWith(HEADER_CAMEL_PREFIX)) {
@@ -232,8 +235,32 @@ public class CamelSinkTask extends SinkTask {
         if (schema.type().equals(Schema.BYTES_SCHEMA.type()) && Objects.equals(schema.name(), Decimal.LOGICAL_NAME)) {
             destination.put(key, Decimal.toLogical(schema, (byte[]) header.value()));
         } else {
-            destination.put(key, header.value());
+            destination.put(key, convertValueFromStruct(header.schema(), header.value()));
         }
+    }
+
+    private static Object convertValueFromStruct(Schema schema, Object value) {
+        // if we have a schema of type Struct, we convert it to map, otherwise
+        // we just return the value as it
+        if (schema != null && value != null && Schema.Type.STRUCT == schema.type()) {
+            return toMap((Struct) value);
+        }
+
+        return value;
+    }
+
+    private static Map<String, Object> toMap(final Struct struct) {
+        final HashMap<String, Object> fieldsToValues = new HashMap<>();
+
+        struct.schema().fields().forEach(field -> {
+            try {
+                fieldsToValues.put(field.name(), struct.get(field));
+            } catch (DataException e) {
+                fieldsToValues.put(field.name(), null);
+            }
+        });
+
+        return fieldsToValues;
     }
 
     CamelKafkaConnectMain getCms() {
--- a/core/src/test/java/org/apache/camel/kafkaconnector/CamelSinkTaskTest.java
+++ b/core/src/test/java/org/apache/camel/kafkaconnector/CamelSinkTaskTest.java
@@ -31,6 +31,7 @@ import org.apache.kafka.connect.data.Decimal;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaAndValue;
 import org.apache.kafka.connect.data.SchemaBuilder;
+import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


oscerd (approved):
Cc @valdar

oscerd (changes requested):
Looking at it a bit better I guess this should be a separated smt in core, which should be enabled/disable through configuration as any other smt. Not all the camel components expect a map as body, so converting each struct to map makes sense only for particular components

omarsmak:
> Looking at it a bit better I guess this should be a separated smt in core, which should be enabled/disable through configuration as any other smt. Not all the camel components expect a map as body, so converting each struct to map makes sense only for particular components

I thought about it at first but then, the Struct itself it is something very specific to Kafka Connect and hence it is pretty useless downstream into Camel components since none will use Struct anyway and the Struct itself it is a Map with schema, then why the extra step to include the SMT since it is not needed. This PR will only convert the Struct to Map if and only if, the Schema is `Struct`, otherwise it will ignore it.

oscerd:
Lets wait for more feedback :-)

lburgazzoli:
With my limited knowledge on the topic (so feel free to correct me if I'm wrong), I think this functionality should really be provided as an an optional SMT. I do agree that none of the components know about Kafka Connect Struct but there's no assurance that a Map is any better so we are just making things a little bit more unpredictable because it will depend on the specific component. 

How Structs should be translated to something the specific sink component is able to digest is - IMHO - something that need to be explicitly configured to avoid unexpected misbehavior.

omarsmak:
> I do agree that none of the components know about Kafka Connect Struct but there's no assurance that a Map is any better so we are just making things a little bit more unpredictable because it will depend on the specific component.

The thing is, `Struct` itself is a decorated Map with the schema support. I understand your concern here but if you want produce data as Map in Kafka Connect, Kafka Connect side of things has only two options, either go with Map or go with Struct in case you want to preserve the schema information, AvroConverter and JsonConverter will produce for you Structs in case the schema information are needed in downstream sinks. However, in our case, Camel components side of things, we don't acknowledge Structs at all and hence, propagating struct vs map downstream has the same effect on downstream component, e.g: If the component expects a String but instead got a Map or Struct. 
However, I'd agree this should be an SMT in case we know there are Camel components that expect Struct but none need that, and hence I couldn't convince myself to ask the user to include a Struct->Map SMT for something we don't even support downstream and hence I just wanted to be automatically converted.

oscerd:
I think a tradeoff would be to make this behavior configurable. Like enabling exactly this code and this execution flow only if an option is set up. Maybe adding it as false as default. I think it sounds like a good tradeoff just in case. @lburgazzoli @valdar @omarsmak

lburgazzoli:
> > I do agree that none of the components know about Kafka Connect Struct but there's no assurance that a Map is any better so we are just making things a little bit more unpredictable because it will depend on the specific component.
> 
> The thing is, `Struct` itself is a decorated Key:Value with the schema support. I understand your concern here but if you want produce data as Map in Kafka Connect, Kafka Connect side of things has only two options, either go with Map or go with Struct in case you want to preserve the schema information, AvroConverter and JsonConverter will produce for you Structs in case the schema information are needed in downstream sinks. However, in our case, Camel components side of things, we don't acknowledge Structs at all and hence, propagating struct vs map downstream has the same effect on downstream component, e.g: If the component expects a String but instead got a Map or Struct.
> However, I'd agree this should be an SMT in case we know there are Camel components that expect Struct but none need that, and hence I couldn't convince myself to ask the user to include a Struct->Map SMT for something we don't even support downstream and hence I just wanted to be automatically converted.

My point is that, as input data is component and external system specific, then you can certainly  convert Struct to a Map and maybe you are lucky but that's really not deterministic. I believe SMT would be better because at this stage you requires to have knowledge of the data shape a component support so to me, it looks like an explicit choice users have to to take. But I'm the least knowledgeable here so I may completely miss the point :)

omarsmak:
> I think a tradeoff would be to make this behavior configurable. Like enabling exactly this code and this execution flow only if an option is set up. Maybe adding it as false as default. I think it sounds like a good tradeoff just in case. @lburgazzoli @valdar @omarsmak

We can add this as an option, as it is a good compromise. We already do this for other part exchange and properties. If others are fine anyway.

omarsmak:
@oscerd @lburgazzoli @valdar I have made it configurable and by default is disable (`false`)

valdar:
I understand the point you are making @omarsmak , still it feels not totally right to me to "hardcode" such behaviour for all components (present and future one) by placing it in `core`. A configurable approach is a decent trade off. Another option would be to have a `typeconverter` (in fact you could just use the one in `camel-debezium-common`) that can be enabled/disabled as needed, in such way, if the needs manifest, we could add other `typeconverters` to handle from struct to something else scenarios.

Moreover there is an ongoing effort to use kamelets in ckc, it is not a trivial change that in my opinion is best to do in steps. Once all the steps are completed we could much more easily change things "per connector" so all of this would make much less sense. In the meantime each addiction to `core` makes the first step toward kamelets usage in ckc little bit more difficult....

omarsmak:
> A configurable approach is a decent trade off.

It is configurable now, but I got what you mean, all what you listed, were in my mind. For example camel-debezium-common dependency which didn't make sense in my head, the SMT option was also in my head but however as I was hesitant due to the headers information and hence I settled on this. Again, I am not convinced of having this as SMT due the reasons I mentioned but if you guys insist on this, I will change this to SMT although now at least is configurable

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
```diff
@@ -225,15 +230,47 @@ If the CamelMainSupport instance fails to be instantiated (ie.: due to missing c
...
 
         if (schema.type().equals(Schema.BYTES_SCHEMA.type()) && Objects.equals(schema.name(), Decimal.LOGICAL_NAME)) {
             destination.put(key, Decimal.toLogical(schema, (byte[]) header.value()));
         } else {
-            destination.put(key, header.value());
+            destination.put(key, convertValueFromStruct(header.schema(), header.value()));
+        }
+    }
+
+    private Object convertValueFromStruct(Schema schema, Object value) {
```
rk3rn3r, line 244: This also converts message keys, so another name like "convertStructToMap()" makes more sense.

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
```diff
@@ -110,7 +114,8 @@ public void start(Map<String, String> props) {
             final String headersRemovePattern = config.getString(CamelSinkConnectorConfig.CAMEL_CONNECTOR_REMOVE_HEADERS_PATTERN_CONF);
             mapProperties = config.getBoolean(CamelSinkConnectorConfig.CAMEL_CONNECTOR_MAP_PROPERTIES_CONF);
             mapHeaders = config.getBoolean(CamelSinkConnectorConfig.CAMEL_CONNECTOR_MAP_HEADERS_CONF);
-            
+            convertStructToMap = config.getBoolean(CamelSinkConnectorConfig.CAMEL_SINK_STRUCT_TO_MAP_CONF);
```
rk3rn3r, line 117: "isConvertStructToMapEnabled" might be a name that better fits the purpose?

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
```diff
@@ -175,8 +180,8 @@ public void put(Collection<SinkRecord> sinkRecords) {
             TaskHelper.logRecordContent(LOG, loggingLevel, record);
 
             Exchange exchange = new DefaultExchange(producer.getCamelContext());
-            exchange.getMessage().setBody(record.value());
-            exchange.getMessage().setHeader(KAFKA_RECORD_KEY_HEADER, record.key());
+            exchange.getMessage().setBody(convertValueFromStruct(record.valueSchema(), record.value()));
```
rk3rn3r, line 183: Is this safe for NPEs on NULL-records which can arrive from SMTs under certain conditions?

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
```diff
@@ -225,15 +230,47 @@ If the CamelMainSupport instance fails to be instantiated (ie.: due to missing c
...
-    private static void mapHeader(Header header, String prefix, Map<String, Object> destination) {
+    private void mapHeader(Header header, String prefix, Map<String, Object> destination) {
         final String key = StringHelper.after(header.key(), prefix, header.key());
         final Schema schema = header.schema();
 
         if (schema.type().equals(Schema.BYTES_SCHEMA.type()) && Objects.equals(schema.name(), Decimal.LOGICAL_NAME)) {
             destination.put(key, Decimal.toLogical(schema, (byte[]) header.value()));
         } else {
-            destination.put(key, header.value());
+            destination.put(key, convertValueFromStruct(header.schema(), header.value()));
```
rk3rn3r, line 240: Does this convert headers to values? I saw somewhere else that there's a way to handover headers (maybe with a prefix?!)

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
```diff
@@ -225,15 +230,47 @@ If the CamelMainSupport instance fails to be instantiated (ie.: due to missing c
...
+            try {
+                Object value = struct.get(field);
+
+                // recursive call if we have nested structs
+                if (value instanceof Struct) {
+                    fieldsToValues.put(field.name(), toMap((Struct) value));
+                } else {
+                    fieldsToValues.put(field.name(), value);
+                }
+            } catch (DataException e) {
```
rk3rn3r, line 268: Isn't DataException only happening on putting content into a Struct but not when adding to a Map?

omarsmak: it is happening when Struct is looking up for the field name, if it doesn't find the field name, it will just throw DataException

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
omarsmak: I think it doesn't but 100% though

omarsmak:
As agreed with @valdar, I will close this PR for now until kamelets lands in KC in order to make things simpler. Once we have it, we will revisit this addition, if would still makes sense or not

rk3rn3r:
Hey everyone! This is a PR that solves the issue I had with the camel-sql-kafka-connector. I left some comments in-line and I think the code can maybe polished a bit, but in general I would like to see that change happening (as you know).

My main point is the architectural view and related developer experience, similar to what Omar already pointed out:

- a Kafka Connect connector can receive very different formats based on the used Kafka Connect Converter (StringConverter or IntegerConverter, etc return String or a respective numeric type; JsonConverter, AvroConverter, ProtobufConverter, etc are schema-ed converters for structural data that yield structural data, mostly a Kafka Connect Struct when a schema is available or a Map when there's no schema available)

- a Kafka Connect *sink* connector is mainly meant to accept data from Kafka and provide it to a downstream system that naturally can't talk to Kafka. The main purpose is converting the standardized input (Map, Struct, String or scalar thing) into the format for the downstream system

- The essence of that PR is that a Kafka Connect Struct is only valid inside Kafka Connect and *should* or must be converted to something that matches the format for downstream processing (SQL for databases, etc). In case of a Camel pipeline from a dev point of view I would either expect there exists a standardized format to handover events between pipeline steps or that ae a generally available format will be used, and for structured/hierarchical data (which is inside a Struct) in Java this would match the Map type.

- That change reliefs also from necessary config changes when the wire format in Kafka Connect is changed (from a schema-less to a schema-ed format): in this case a user needs to update the pipeline and inlcude the proposed SMT which was not needed prior the wire format change. The basic idea of a serialization format in Kafka Connect is quite decoupled from the data that are handled and send to the downstream system/s. When you pass down the Struct to downstream systems then you imo force handling Kafka internals in the downstream system/connector.

core/src/main/java/org/apache/camel/kafkaconnector/CamelSinkTask.java
omarsmak: aha actually is not needed since field lookup is on the bases of the `Field`, not the name. Good catch!

PR #1079, opened 2021-03-03
Fix Netty tests reliability

orpiske:
- avoid per class lifecycle since it seems to cause the code to try to bind the same port multiple times
- do not assume local hostnames and/or that they are correctly configured


oscerd (approved):

tadayosi:
@orpiske Ah awesome. Thanks!

tadayosi (approved):

Issue #1077, opened 2021-03-02
HTTPS-sink-connector duplicates protocol 


OlyaVashchenko:
Hello. I'm trying to use **camel-https-kafka-connector** to _POST_ data into _SuccessFactors_ (with basic authentication - username and password) and get the following **error in my console**: 
```
Caused by: org.apache.camel.ResolveEndpointFailedException: Failed to resolve endpoint: https://https:apisalesdemo8.successfactors.com/odata/v2/upsert?httpMethod=POST due to: The uri part is not configured correctly. You have duplicated the http(s) protocol.
        at org.apache.camel.component.http.HttpComponent.createEndpoint(HttpComponent.java:298)
        at org.apache.camel.support.DefaultComponent.createEndpoint(DefaultComponent.java:171)
        at org.apache.camel.impl.engine.AbstractCamelContext.doGetEndpoint(AbstractCamelContext.java:920)
```
My **.properties** file:
```
name=CamelHttpsSinkConnector
connector.class=org.apache.camel.kafkaconnector.https.CamelHttpsSinkConnector
tasks.max=1
key.converter=org.apache.kafka.connect.storage.StringConverter
value.converter=org.apache.kafka.connect.storage.StringConverter
topics=camel-test
camel.sink.path.httpUri=apisalesdemo8.successfactors.com/odata/v2/upsert?processInactiveEmployees=true&purgeType=full
camel.sink.endpoint.httpMethod=POST
#camel.sink.endpoint.authMethod=Basic
#camel.sink.endpoint.authUsername=xxx
#camel.sink.endpoint.authPassword=xxx
```
If I uncomment the authentication part, I get an error of the same nature.
Kafka version: 2.13-2.7.0, connector version: 0.8.0

OlyaVashchenko:
Just in case: I downloaded the connector from **https://camel.apache.org/camel-kafka-connector/latest/connectors.html**

tadayosi:
Looks like this is another issue that EndpointUriFactory causes.

oscerd:
@tadayosi since you worked on other feel free to have a look at this one too. I think the scheme is wrong, but the duplication seems to be a different problem.

tadayosi:
As demonstrated in my itests, the issue occurs when `httpUri` includes additional query string such as `?processInactiveEmployees=true&purgeType=full`. I believe it requires camel-core to be patched.

A quick workaround for the time being would be to use `camel.sink.url` to provide the entire URL yourself.

[...]

Issue #1085, opened 2021-03-05
netty-http sink itests - mock web server out of test code into JUnit 5 extension


tadayosi:
Follow up #1084

PR #1096, opened 2021-03-11
Add https itests

tadayosi:
Tried to reproduce #1077 with the itests but I couldn't.


tadayosi:
Will work on #1085 in the next pull req.

tadayosi:
OK, I now reproduced #1077 with the itests.

oscerd (approved):

Issue #1147, opened 2021-04-20
Jackson conflicts causing Azure Storage Blob tests to fail


orpiske:
Our tests are failing because it is are carrying over Jackson 2.12.x from another component and the Azure SDK is known to have issues with 2.12.x [1]. We'll probably be fine once we upgrade to a version of the Azure SDK newer than 12.10 on Camel Core, but for now we'll have to override locally.

1. https://github.com/Azure/azure-sdk-for-java/issues/9465
2. https://github.com/Azure/azure-sdk-for-java/issues/19392

orpiske:
Fixed.

PR #1150, opened 2021-04-21
Avoid using canonical host due to multiple address/port allocation issues on GH actions

orpiske:


oscerd (approved):

orpiske:
The test failed on Azure, which is related to #1147. So, let's merge this and see how it goes with that one.

PR #1160, opened 2021-04-28
Minor fixes for AWS SQS tests when running against actual AWS

orpiske:


oscerd (approved):

orpiske:
The test passed in 2 of 3 of the checks. It's stuck in one of them, but it doesn't seem relevant to the change. I'll rollback if broken. Therefore, merging.

PR #1175, opened 2021-05-14
Fix AWS2 Kinesis Sink issues

orpiske:
AWS2 Kinesis Sink test was failing because the shard was not available
at the time the check was occurring


oscerd (approved):

valdar (commented):
:+1:

PR #1199, opened 2021-06-08
Some documentation link fixes.

valdar:


oscerd (approved):

valdar:
Should be in `main` then when all is as expected I will beckport

PR #1198, opened 2021-06-07
Fixed [camel-kafka-connector-docs-maven-plugin] : make some parameters configurable

abhikarkamkar:
Fix for [#1195](https://github.com/apache/camel-kafka-connector/issues/1195)
Made following parameters configurable:-
- SINK_CONNECTOR_LINK_SUFFIX_ADOC
- SOURCE_CONNECTOR_LINK_SUFFIX_ADOC
- XREF_CONNECTOR_LINK_PREFIX
- DOC_FOLDER_WEBSITE
- PROJECT_BASE_DIR

Issue #1195, opened 2021-06-07
[camel-kafka-connector-docs-maven-plugin] : make some parameters configurable


valdar:
Please see:
https://github.com/apache/camel-kafka-connector/blob/main/tooling/camel-kafka-connector-docs-maven-plugin/src/main/java/org/apache/camel/kafkaconnector/maven/docs/UpdateDocComponentsListMojo.java#L51
https://github.com/apache/camel-kafka-connector/blob/main/tooling/camel-kafka-connector-docs-maven-plugin/src/main/java/org/apache/camel/kafkaconnector/maven/docs/UpdateDocComponentsListMojo.java#L171

abhikarkamkar:
@valdar can I take this

abhikarkamkar:
Hi can I work on this

PR #1206, opened 2021-06-12
Fixed [camel-kafka-connector-docs-maven-plugin] : make some parameters configurable #1198

abhikarkamkar:
Fix for #1195
Made following parameters as maven plugin parameters:-

SINK_CONNECTOR_LINK_SUFFIX_ADOC
SOURCE_CONNECTOR_LINK_SUFFIX_ADOC
XREF_CONNECTOR_LINK_PREFIX
DOC_FOLDER_WEBSITE
PROJECT_BASE_DIR


tooling/camel-kafka-connector-docs-maven-plugin/pom.xml
```diff
@@ -158,4 +158,20 @@
         </dependency>
     </dependencies>
 
+    <build>
```
oscerd, line 161: You don't need to do this in tooling. There is the MOJO invocation in connectors/pom.xml and you should set the option there eventually, but they already have default value, so this seems to be useless

oscerd (changes requested):
I think this needs more work

PR #1316, opened 2022-01-15
camel-website #701: RI info (1.0.x)

djencks:
See apache/camel-website#701

Note that:
- the vote for 1.0.0 has not yet completed. Presumably it should complete before this is applied?
- This uses the Antora version 1.0.x.  I suggest we change the 0.11.x Antora version from 0.11.0 to 0.11.x similarly.
- The vote doesn't indicate this is LTS, so I said it wasn't.

### DIFF
--- a/connectors/antora.yml
+++ b/connectors/antora.yml
@@ -1,5 +1,21 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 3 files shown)


valdar:
The LTS is based on the camel version, so if the camel version we depend upon is an LTS so it is the corresponding ckc release.

So in this specific case, both `0.11.0` and `1.0.0` are LTS releases.

I guess we can explicitly mention it in the votes.

oscerd (approved):

PR #1344, opened 2022-03-03
docs: reference jars in version 1.0.0 instead of 0.11.0 in documentation

CrowdSalat:


oscerd (changes requested):
This is autogenerated as far as I remember. So it should be done somewhere else not by hands

oscerd:
Fixed with #1348 and #1350

PR #1366, opened 2022-05-06
Correcting a Kafka terminology in Camel Kafka doc

valathsu:
Apache Kafka uses source/sink to define source/target systems. Looks like 'sink' is mistakenly written as 'syncs' in this doc.


valathsu:
will raise from a new repo

PR #1405, opened 2022-07-27
(chores) Multiple test fixes and improvements

orpiske:
- Align test containers version with the same one used by Camel Core
- Fix test issues on Apple Silicon


orpiske:
AWS 2 ones will be on a separate PR.

oscerd (approved):
We should also try to switch to 3.18.0

orpiske:
> We should also try to switch to 3.18.0

+1. I'll try to do that after I get the tests in good shape.

PR #1450, opened 2022-10-17
Align main branch to Camel 3.19.0

oscerd:


oscerd:
Let's aling and then work on IT Tests

PR #1483, opened 2022-12-07
Added exlusion of kamelets generation.

valdar:


connectors/pom.xml
```diff
@@ -104,6 +104,9 @@
                     <exampleSinkPropertiesFileTemplate>${project.basedir}/camel-kafka-connector-template-example-sink-properties.template</exampleSinkPropertiesFileTemplate>
                     <!-- WARNING: connectors not in catalog will be removed -->
                     <removeMissingComponents>true</removeMissingComponents>
+                    <exludedKamelets>
```
oscerd, line 107: I think the placeholder is wrong. It's excludedKamelets looking at the MOJO

oscerd (changes requested):
Just a minor finding.

connectors/pom.xml
valdar: good catch tnx!

PR #1503, opened 2023-01-20
feat: add bulk upsert connector

Matheusafonsouza:


oscerd (changes requested):
Thanks, but you just need to add the kamelet to the kamelets repository and once we upgrade the version the connector will be automatically generated.

Matheusafonsouza:
Thanks, i'll be doing It after make some tests on the connector

valdar:
@Matheusafonsouza tnx for the contribution. Let me know if you need some help/guidance to contribute this in camel kamelets project.

Matheusafonsouza:
@valdar i made some tests and the connector did not worked, i think that`s because i dont know much about the payload, no errors rised, just nothing

PR #1628, opened 2024-05-08
Upgrade to kafka 3.7.0

davsclaus:


davsclaus:
we should use the kafka version that camel 4.4.x uses

davsclaus:
it should be 3.6.1

PR #1640, opened 2024-05-19
Filter camel properties to camelMain

seanshi007:
Previously all properties are first put into camelProperties and then logged (after filter out sensitive keys defined in Camel core). When producer or consumer is overridden, e.g. when "producer.override.sasl.jaas.config" is set, which will not be filtered out and hence being logged and causing potential secret exposure. 

This change will filter out non-camel properties as they will not be used by camel core anyway.

### DIFF
--- a/core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectMain.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectMain.java
@@ -47,6 +47,8 @@ public class CamelKafkaConnectMain extends SimpleMain {
     public static final String KAMELET_IDEMPOTENT_TEMPLATE_PARAMETERS_PREFIX = "camel.kamelet.ckcIdempotent.";
     public static final String KAMELET_REMOVEHEADER_TEMPLATE_PARAMETERS_PREFIX = "camel.kamelet.ckcRemoveHeader.";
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectMain.java
```diff
@@ -227,7 +229,10 @@ public CamelKafkaConnectMain build(CamelContext camelContext) {
             camelMain.configure().setDumpRoutes(Boolean.TRUE.toString());
 
             Properties camelProperties = new Properties();
-            camelProperties.putAll(props);
+            Map<String, String> camelProps = props.entrySet().stream()
```
valdar, line 232: The problem here is that, theoretically (since it is not explicitly tested in tests, would be possible to add properties not starting with `camel.` as a way to configure some other useful stuff like factories for some camel components i.e. as in https://github.com/apache/camel-kafka-connector/blob/main/connectors/camel-kafka-connector-fix-dependencies.properties#L25 what about having a specific property say `camel.kafkaconnector.filtered.props` as a comma separated list of sensitive properties that one may want to explicitly filter?

valdar (changes requested):
First of all thanks a lot for the contribution!
Please see my inline comments.

davsclaus:
closing old PR that is not active

PR #1671, opened 2024-09-17
Introduce truststore password configuration for kafka-ssl-sink

XLAQO:
Hi,

I was trying to use the Kafka SSL Sink with a truststore. Of course, truststores usually have a password set, which cannot be configured for this connector. This Pull Request aims to add this configuration option.


oscerd (changes requested):
This will be overriden on each build, it will change once we release the catalog on the Kamelet side. Closing, but thanks

Issue #1798, opened 2026-08-24
Five connector configs declare password options as ConfigDef.Type.STRING


oscerd:
## Description

The kamelet-based generator emits `ConfigDef.Type.PASSWORD` when a kamelet property declares
`format: password` (`CamelKafkaConnectorKameletUpdateMojo.addConnectorOptions`, lines 606-611).
Across the shipped connector modules this holds for the overwhelming majority of password/passphrase
options, but five checked-in config classes declare them as `ConfigDef.Type.STRING` even though the
kamelet YAML bundled in the same module declares `format: password` (and
`x-descriptors: urn:camel:group:credentials`):

| Connector module | Option | Current type |
|---|---|---|
| `camel-jms-pooled-apache-artemis-sink-kafka-connector` | `camel.kamelet.jms-pooled-apache-artemis-sink.password` | `STRING` |
| `camel-jms-pooled-apache-artemis-source-kafka-connector` | `camel.kamelet.jms-pooled-apache-artemis-source.password` | `STRING` |
| `camel-scp-sink-kafka-connector` | `camel.kamelet.scp-sink.privateKeyPassphrase` | `STRING` |
| `camel-sftp-sink-kafka-connector` | `camel.kamelet.sftp-sink.privateKeyPassphrase` | `STRING` |
| `camel-sftp-source-kafka-connector` | `camel.kamelet.sftp-source.privateKeyPassphrase` | `STRING` |

`camel-sftp-sink-kafka-connector` is a good illustration — in the same file, `password` is
`Type.PASSWORD` (line 87) while `privateKeyPassphrase` is `Type.STRING` (line 93), although both are
`format: password` in `sftp-sink.kamelet.yaml`.

These look like generated output that predates the current generator behaviour and was never
regenerated.

## Expected Behavior

Every connector option whose kamelet declares `format: password` (or whose Camel component metadata
declares `secret: true`) is `ConfigDef.Type.PASSWORD` in the shipped `Camel*ConnectorConfig` class,
so that Kafka Connect applies its usual handling for password-typed configuration.

## Actual Behavior

Five options are `Type.STRING` and therefore do not get that handling, inconsistently with the ~270
equivalent options in the rest of the connector set.

## Additional Context

Two parts:

1. Correct the five modules listed above.
2. Close the drift so it cannot recur:
   - the kamelet path decides `PASSWORD` solely on `"password".equals(kameletProperty.getFormat())`;
[...]

Issue #423, opened 2020-09-04
Investigate using Kamelet concepts in Camel-Kafka-connector


oscerd:

lburgazzoli:
I think Kamelets or a very similar concept can also be  used to implement Kafka Connect SMTs which could supersede #22 and #21  by unleashing the full power of the Camel DSL

oscerd:
Lets discuss here. What we know so far:
1. The kamelet should an alternative approach to the current approach
2. the kamelet yaml should be packaged with the connector
3. We should re-use the itest infra we have

@lburgazzoli @valdar

heiko-braun:
Should Kamelets be an alternative, or a replacement for the current approach?

valdar:
@heiko-braun as first step a drop in replacement then transitioning to a replacement.

PR #1808, opened 2026-08-24
Fix #1795: ckcUnMarshal route template applies unmarshal instead of marshal

### LINKED ISSUE #1795: [BUG] CamelSyslogSourceConnector fails with NoTypeConversionAvailableException due to marshal/unmarshal typo in ckcUnMarshal route template
## Description
`CamelSyslogSourceConnector` fails when receiving syslog messages with the following error:

```
org.apache.camel.NoTypeConversionAvailableException: No type converter available to 
convert from type: io.netty.buffer.PooledUnsafeDirectByteBuf to the required type: 
org.apache.camel.component.syslog.SyslogMessage
```

## Version
- camel-kafka-connector: 4.18.0
- Apache Camel: 4.18.0
- Kafka: 4.3.1

## Root Cause
In `CamelKafkaConnectMain.java`, the `ckcUnMarshal` route template incorrectly calls 
`.marshal()` instead of `.unmarshal()`:

```java
// Bug - in core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectMain.java
routeTemplate("ckcUnMarshal")
        .templateParameter("unmarshal", "dummyDataformat")
        .from("kamelet:source")
        .marshal("{{unmarshal}}")   // <- should be .unmarshal()
        .to("kamelet:sink");
```

This causes the raw Netty `PooledUnsafeDirectByteBuf` to never get deserialized into 
a `SyslogMessage`, resulting in the error above.

This is a regression from `0.11.x` where the route was built directly without kamelets:

```java
// 0.11.5 - worked correctly
rd.unmarshal(unmarshallDataFormat);
```

## Fix
Change `.marshal()` to `.unmarshal()` in the `ckcUnMarshal` route template:

```java
routeTemplate("ckcUnMarshal")
        .templateParameter("unmarshal", "dummyDataformat")
        .from("kamelet:source")
        .unmarshal("{{unmarshal}}")   // <- fixed
        .to("kamelet:sink");
```

## Impact
This bug affects any connector that uses `camel.source.unmarshal`, not just syslog.

## Steps to Reproduce
1. Set up Kafka Connect with `camel-syslog-kafka-connector` version `4.18.0`
2. Configure `CamelSyslogSourceConnector` to listen on UDP port 514
3. Send a syslog message
4. Observe `NoTypeConversionAvailableException` in the logs

## Expected Behavior
Syslog messages are correctly unmarshalled into `SyslogMessage` and published to Kafka topic.

oscerd:
Fixes #1795.

## What

The `ckcUnMarshal` route template was calling `.marshal("{{unmarshal}}")`:

```java
routeTemplate("ckcUnMarshal")
        .templateParameter("unmarshal", "dummyDataformat")
        .from("kamelet:source")
        .marshal("{{unmarshal}}")   // <- applied the wrong direction
        .to("kamelet:sink");
```

so every connector configured with `camel.sink.unmarshal` or `camel.source.unmarshal` got a route
that marshalled the payload instead of unmarshalling it. The reporter hit this as a
`NoTypeConversionAvailableException` on `CamelSyslogSourceConnector`, but it affects every use of
either option.

## Why it regressed

`ca4e784352` ("Related to #423 modularized kamelets and composed them to better autogenerate
connectors from kamelets catalog") extracted the marshal and unmarshal steps into standalone route
templates. Before that commit the route was built directly and was correct:

```java
rdInTemplateSource = rdInTemplateSource.unmarshal("{{unmarshal}}");
```

The new `ckcUnMarshal` template was copied from `ckcMarshal` directly above it and the operation was
never switched over. This PR restores the pre-`ca4e784352` behaviour.

## Test coverage

The existing `DataFormatTest` cases only assert that the context starts and that the data format
resolves — never which direction it is applied in — which is why this passed CI for so long.

Added two tests that push a message through the built route with a data format that records which
direction it was invoked in:

- `testUnmarshalDataFormatIsAppliedInTheUnmarshalDirection`
- `testMarshalDataFormatIsAppliedInTheMarshalDirection`

Confirmed that the first one fails on the unpatched template (`expected: <true> but was: <false>`)
and passes with the fix, so it is a real regression guard.

## Verification

- `core` module: 103/103 tests pass.
- Full reactor build from the repository root (`./mvnw clean install`, all tests): BUILD SUCCESS.

Note the root build also surfaced pre-existing regeneration drift in the checked-in connector tree
(unrelated to this change, and left out of this PR) — filed separately as #1798.

### DIFF
--- a/core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectMain.java
+++ b/core/src/main/java/org/apache/camel/kafkaconnector/utils/CamelKafkaConnectMain.java
@@ -316,7 +316,7 @@ public class CamelKafkaConnectMain extends SimpleMain {
                     routeTemplate("ckcUnMarshal")
                             .templateParameter("unmarshal", "dummyDataformat")
                             .from("kamelet:source")
-                            .marshal("{{unmarshal}}")
+                            .unmarshal("{{unmarshal}}")
                             .to("kamelet:sink");
 
                     //create aggregator template
--- a/core/src/test/java/org/apache/camel/kafkaconnector/DataFormatTest.java
+++ b/core/src/test/java/org/apache/camel/kafkaconnector/DataFormatTest.java
@@ -16,19 +16,25 @@
  */
 package org.apache.camel.kafkaconnector;
 
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.camel.Exchange;
 import org.apache.camel.component.hl7.HL7DataFormat;
 import org.apache.camel.component.syslog.SyslogDataFormat;
 import org.apache.camel.impl.DefaultCamelContext;
 import org.apache.camel.kafkaconnector.utils.CamelKafkaConnectMain;
+import org.apache.camel.spi.DataFormat;
+import org.apache.camel.support.service.ServiceSupport;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class DataFormatTest {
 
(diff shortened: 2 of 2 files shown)


oscerd:
Superseded by #1796, which was merged while this was open and makes the identical one-line change to
the `ckcUnMarshal` template, plus a syslog-based unmarshal-direction test. Closing in favour of it —
apologies for the duplicate effort, @adityaanikam got there first.

One gap worth keeping: `main` now guards the unmarshal direction, but nothing asserts that
`ckcMarshal` still marshals, so the symmetric mistake in the other template would pass CI. I'll open a
small follow-up with just that test rather than re-propose this branch.

Note also that #1796's merge left the new imports unsorted, which fails checkstyle under
`-Psourcecheck` and is currently making `main` red — fix in #1813.

